PR #1, opened 2021-09-28
README.md: adds repo description

DnPlas:
Adds repository description.


metadata.yaml
```diff
@@ -0,0 +1,25 @@
+# Copyright 2021 Daniela Plascencia
+# See LICENSE file for licensing details.
+
+# For a complete list of supported options, see:
```
knkski, line 4: We can probably also just remove this comment

metadata.yaml
```diff
@@ -0,0 +1,25 @@
+# Copyright 2021 Daniela Plascencia
```
knkski, line 1: I think the copyright notice should be for Canonical? Maybe @evilnick knows

metadata.yaml
DnPlas: Oops, didn't notice it. Will remove it.

metadata.yaml
DnPlas: Those are inherited from 'charmcraft init', will edit/remove them.

charmcraft.yaml
```diff
@@ -0,0 +1,19 @@
+# Learn more about charmcraft.yaml configuration at:
```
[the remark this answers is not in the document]
DnPlas: Yes, same as below.

README.md
```diff
@@ -1,2 +1,13 @@
...
-Kubeflow Training Operator
+# Training Operator
+
+## Overview
+
+This repository hosts the Kubernetes Training Operator for Kubeflow training jobs.
+
+## Description
+
+The [Kubeflow Training Operator](https://www.kubeflow.org/docs/components/training/) provides Kubernetes custom resources to run distributed or non-distributed training through so called training jobs, such as TFJobs and PytrochJobs. The Training Operator in this repository is a Python script that wraps the latest released Kubeflow Training Operator [manifests](https://github.com/kubeflow/manifests/tree/master/apps/training-operator), providing lifecycle management and handling events (install, upgrade, integrate, remove). It is part of the [Kubeflow bundle](https://github.com/canonical/bundle-kubeflow) in the Juju charm ecosystem.
```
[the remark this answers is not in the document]
DnPlas: +1 on the reference-style links. I'll re-phrase that sentence and correct the typo.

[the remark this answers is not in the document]
evilnick: Yes to adding linebreaks, even if they are manual and inconsistent. 99% of good markdown engines don't care about a single LF and will easily render it properly. The main benefit is that it makes reviews much easier

README.md
```diff
@@ -1,2 +1,18 @@
...
-Kubeflow Training Operator
+# Training Operator
+
+## Overview
+
+This repository hosts the Kubernetes Training Operator for Kubeflow training jobs.
+
+## Description
+
+The [Kubeflow Training Operator][1] provides Kubernetes custom resources to run distributed or non-distributed training through so called training jobs, such as TFJobs and PytorchJobs. The Training Operator in this repository is a Python script that wraps the latest released Kubeflow Training Operator [manifests][2], providing lifecycle management and handling events (install, upgrade, integrate, remove). It is part of the [Charmed Kubeflow][3] operators.
```
evilnick, line 9: I don't like "so called", it is usually used in a derogatory way.  I think it works fine if you just drop "training through so called"

evilnick, line 9: Also, since you are editing it anyhow, it is more correct to say `Python script which wraps...`

README.md
```diff
@@ -1,2 +1,18 @@
...
+
+## Usage
+
+While it is possible to deploy the Training Operator as a standalone operator, it works best when deployed alongside other components included in the Kubeflow bundle. For installation steps, please refer to the [installation][4] guide.
+
+[1]: https://www.kubeflow.org/docs/components/training/
```
evilnick, line 15: These are fine. Usually in docs we use the implicit reference.  so we would put `[docs][]` in the text and then the reference is just `[docs]: http://whatever`. It hardly matters here, but in larger docs it makes it much easier to correlate the links when/if they need updating

README.md
evilnick, line 9: shouldn't it be `one of the  [Charmed Kubeflow][3] operators.`?

DnPlas: +1

README.md
DnPlas: Noted

README.md
DnPlas: Done

metadata.yaml
```diff
@@ -0,0 +1,19 @@
...
+maintainers: [Juju Developers <juju@lists.ubuntu.com>]
+tags: [ai, bigdata, kubeflow, machine-learning, tensorflow, pytorch]
+containers:
+  training-operator:
+    resource: training-operator-image
+resources:
+  training-operator-image:
+    type: oci-image
+    description: OCI image for training-operator
+    auto-fetch: true
```
jnsgruk, line 18: This field is deprecated, technically so is `upstream-source`, though there are benefits to including that for testing purposes.

```suggestion
```

metadata.yaml
```diff
@@ -0,0 +1,19 @@
+name: training-operator
+display-name: Training Operator
+summary: Tools for distributed or non-distributed training on Kubernetes
+description: |
+  Training Operator provides Kubernetes custom resources that make it
+  easy to run distributed or non-distributed training jobs on Kubernetes.
+  This charm deploys the Operator configured for use with Kubeflow to
+  Kubernetes models in Juju.
+maintainers: [Juju Developers <juju@lists.ubuntu.com>]
+tags: [ai, bigdata, kubeflow, machine-learning, tensorflow, pytorch]
```
jnsgruk, line 10: This is also essentially a no-op for Charmhub -- this field used to be read by the old charm store.

```suggestion
```

metadata.yaml
```diff
@@ -0,0 +1,19 @@
+name: training-operator
+display-name: Training Operator
+summary: Tools for distributed or non-distributed training on Kubernetes
+description: |
+  Training Operator provides Kubernetes custom resources that make it
+  easy to run distributed or non-distributed training jobs on Kubernetes.
+  This charm deploys the Operator configured for use with Kubeflow to
+  Kubernetes models in Juju.
+maintainers: [Juju Developers <juju@lists.ubuntu.com>]
```
jnsgruk, line 9: I'd prefer for this to include an actual name; this makes it look like the Juju team is maintaining this charm :)

jnsgruk (commented):
Good start, look forward to seeing the rest ;-)

metadata.yaml
DnPlas: I'll remove both keys from now and add `upstream-source` later if I need it for testing.

metadata.yaml
DnPlas: +1 on the suggestion, thanks

metadata.yaml
DnPlas: @knkski @DomFleischmann what do you guys think of this?

DomFleischmann: There is currently a discussion about this maintainers field. I do agree that Juju Developers doesn't really make sense and I'm not sure if I'm even subscribed to this mailing list.

I personally feel like you can just add your own name and canonical email address there, that seems to be the direction charmhub is going anyways.

jnsgruk: In general, my preference is that there is an actual name associated with charms on Charmhub. That doesn't mean that the Kubeflow team can't be added as a contributor on the backend so you can all push/upload/release etc., but gives a more direct sense that it's 'owned'

metadata.yaml
knkski: `upstream-source` is in a weird state that I wouldn't call deprecated exactly. I opened this bug to talk about adding official Juju support for this or a similar field:

https://bugs.launchpad.net/juju/+bug/1946121

At any rate, for Kubeflow charms, a lot of our tooling uses `upstream-source`, so we should probably keep it in for now.

DnPlas: Alright, keeping it then.

knkski (approved):
LGTM

metadata.yaml
DnPlas: I updated it with my name and email address. Thanks.

PR #2, opened 2021-10-05
Adds src/charm.py and config.yaml with initial config

DnPlas:
Besides creating the files, adds the pebble layer configuration and initial config schema for the training operator.

PR #3, opened 2021-10-05
Adds on_config_changed event handler and runtime deps

DnPlas:
Please note this is a draft and because it depends on #1 and #2 , it shouldn't be merged before them. As in #2 , this PR's target branch is `dnplas/1.4-dev` since `main` is still empty. Once we start populating the repository, we can merge on `main`. Your comments and edit suggestions are welcomed. I'll work on applicable changes in my fork.

For this PR, let's just focus on commits: 
def836b Adds requirements.txt and enables git in charmcraft.yaml
deb14d7 src/charm.py: adds _on_config_changed event handler

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,1 @@
+git+https://github.com/canonical/operator/#egg=ops
--- a/src/charm.py
+++ b/src/charm.py
@@ -22,7 +22,7 @@ class TrainingOperatorCharm(CharmBase):
         self._namespace = self.model.name
         self._manager_service = "manager"
 
-        self.framework.observe(self.on.install, self._on_install)
+        self.framework.observe(self.on.install, self._on_config_changed)
         self.framework.observe(self.on.config_changed, self._on_config_changed)
         self.framework.observe(self.on.training_operator_pebble_ready, self._on_training_operator_pebble_ready)
 
@@ -61,9 +61,26 @@ class TrainingOperatorCharm(CharmBase):
 
     def _on_config_changed(self, event):
         """
-        Placeholder for config changed event
+        Manages the confg changed event and will only take effect on the
+        manager service.
+        This piece of software creates a new Pebble configuration layer,
+        updates the Pebble plan, and restarts the manager service.
         """
-        pass
+        container = self.unit.get_container(self._name)
+        # Get current config
+        current_layer = container.get_plan().services
+        # Create a new config layer
+        new_layer = self._training_operator_layer()
+        if container.can_connect():
+            # Check if there are any changes
+            container.add_layer(self._manager_service, new_layer, combine=True)
+            if current_layer != new_layer["services"]:
+                logging.info("Updated Pebble plan with new configuration")
+                container.restart(self._manager_service)
+                logging.info("Restart training-operator")
+            self.unit.status = ActiveStatus()
+        else:
+            self.unit.status = WaitingStatus("Waiting for Pebble in workload container")
 
     def _on_install(self, event):
         """
--- a/src/charm.py
+++ b/src/charm.py
@@ -1,0 +1,75 @@
+#!/usr/bin/env python3
+
+import logging
+
+from ops.main import main
+from ops.pebble import Layer
+from ops.charm import CharmBase
+from ops.model import ActiveStatus, WaitingStatus, MaintenanceStatus
+
+logger = logging.getLogger(__name__)
+
+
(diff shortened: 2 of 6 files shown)


DomFleischmann:
I don't mind multiple PRs while developing a Charm but I think changes in the same file like `src/charm.py` are better kept in one. Else it gets a bit confusing to see which one comes first.

I would propose you move these changes to #2

requirements.txt
```diff
@@ -0,0 +1 @@
+git+https://github.com/canonical/operator/#egg=ops
```
DomFleischmann, line 1: This package is on [pypi](https://pypi.org/project/ops/) so you shouldn't have to pull it from github

jnsgruk: In this case, @DnPlas is using `container.can_connect` which hasn't yet made it into a release, so this is necessary :)

src/charm.py
```diff
@@ -0,0 +1,92 @@
...
+
+        self._name = self.model.app.name
+        self._namespace = self.model.name
+        self._manager_service = "manager"
+
+        self.framework.observe(self.on.install, self._on_config_changed)
+        self.framework.observe(self.on.config_changed, self._on_config_changed)
+        self.framework.observe(self.on.training_operator_pebble_ready, self._on_training_operator_pebble_ready)
+
+    def _training_operator_layer(self):
```
jnsgruk, line 29: This could probably use the `@property` decorator
```suggestion
    @property
    def _training_operator_layer(self) -> dict:
```

src/charm.py
```diff
@@ -0,0 +1,92 @@
...
+        """
+        Returns a Pebble configuration layer for Kubeflow Training Operator
+        This piece of software takes care of the deployment part of the
+        training-operator
+        """
```
jnsgruk, line 34: We generally use the Google docstring format for operator framework and charms. This docstring might look like the following using that format:

```suggestion
        """Property that returns a pre-configured Pebble layer.
        
        The Pebble layer contains information about how to start the manager service,
        and sets the relevant environment variables for the process to start successfully.
        
        Returns:
            dict: a dictionary representation of a Pebble layer
        """
```

Note:
    - First line is on the same line as the opening `"""`, ends with a full stop
    - The second section allows you to add more context if needed (though probably isn't here)
    - There should be seperate sections for `Args`, `Raises`, `Returns`.

There are a bunch more examples here: https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

src/charm.py
```diff
@@ -0,0 +1,92 @@
...
+    def _on_config_changed(self, event):
+        """
+        Manages the confg changed event and will only take effect on the
+        manager service.
+        This piece of software creates a new Pebble configuration layer,
+        updates the Pebble plan, and restarts the manager service.
+        """
+        container = self.unit.get_container(self._name)
+        # Get current config
+        current_layer = container.get_plan().services
```
jnsgruk, line 71: Did you mean: `container.get_services()` or was there a different purpose for grabbing the whole plan? :)

DnPlas:
> I don't mind multiple PRs while developing a Charm but I think changes in the same file like `src/charm.py` are better kept in one. Else it gets a bit confusing to see which one comes first.
> 
> I would propose you move these changes to #2

@DomFleischmann Please let me know if you still think this is necessary.

requirements.txt
DnPlas: Basically that's the reason. Resolving this.

src/charm.py
DnPlas: +1, will use it from now on.

DomFleischmann:
> > I don't mind multiple PRs while developing a Charm but I think changes in the same file like `src/charm.py` are better kept in one. Else it gets a bit confusing to see which one comes first.
> > I would propose you move these changes to #2
> 
> @DomFleischmann Please let me know if you still think this is necessary.

Let's keep it how is and see how it goes.

src/charm.py
DnPlas: Is there a preferred method for creating Pebble layers? In the sense that, I could leave it as a dict or use [this](https://github.com/canonical/operator/blob/1.2.0/ops/pebble.py#L445) to instantiate and use the attributes later on, like `service` for this particular charm.

jnsgruk: Either is fine, using `Layer` is probably preferable, just annotate the return type accordingly

PR #5, opened 2021-10-11
Enables on-install event handler

DnPlas:
The on-install event handler creates all needed k8s resources for the training-operator using the [lightkube](https://github.com/gtsystem/lightkube).

This PR introduces the following changes:
* src/charm.py: adds BlockedStatus logic
* metadata.yaml: updates upstream-source image
* src/: enables crds creation and adds crds_manifests.yaml
* src/: enables auth resources creation and adds auth_manifests.yaml
* requirements.txt: Adds lightkube to runtime dependencies

> NOTE: in the previous version of this PR, these changes were done using the kubernetes python client. Moving away from that in current version.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -2,12 +2,13 @@
 
 import os
 import logging
+import traceback
 from pathlib import Path
 
 from ops.main import main
 from ops.pebble import Layer
 from ops.charm import CharmBase
-from ops.model import ActiveStatus, MaintenanceStatus, WaitingStatus
+from ops.model import ActiveStatus, BlockedStatus, MaintenanceStatus, WaitingStatus
 
 from lightkube import ApiError, Client, codecs
 from lightkube.types import PatchType
@@ -134,13 +135,17 @@ class TrainingOperatorCharm(CharmBase):
 
     def _on_install(self, _):
         """Event handler for on-install events."""
-        auth = self._patch_auth_resources()
-        crds = self._create_crds()
-
-        # Check resources were applied/created correctly
-        if auth and crds:
+        try:
+            self._patch_auth_resources()
+            self._create_crds()
+        except ApiError as e:
+            logging.error(f"Some Kubernetes resources failed to be created/patched")
+            logging.error(traceback.format_exc())
+            self.unit.status = BlockedStatus(
+                f"Creating/patching resources failed with code {str(e.status.code)}."
+            )
+        else:
             self.unit.status = ActiveStatus()
-
 
 if __name__ == "__main__":
     main(TrainingOperatorCharm)


src/charm.py
```diff
@@ -80,10 +82,52 @@ def _on_config_changed(self, _):
...
+        self.unit.status = MaintenanceStatus("Creating ServiceAccount")
+        try:
+            K8sResources.create_service_account()
+        except ResourceHandlingFailed as e:
+            logging.error(
+                "Unable to create ServiceAccount for training-operator: %s", str(e)
+            )
+        else:
+            logging.info("Successfully created ServiceAccount for training-operator")
+            self.unit.status = ActiveStatus()
```
jnsgruk, line 96: I would omit this from this particular method. I think it's better off after the two calls to `_create_service_account` and `_create_cluster_role_binding` in your `_on_install` method.

My preference is to keep any status changes within the event handlers themselves, rather than in helper methods :)

src/charm.py
```diff
@@ -80,10 +82,52 @@ def _on_config_changed(self, _):
...
+        self.unit.status = MaintenanceStatus("Creating ClusterRole")
+        try:
+            K8sResources.create_cluster_role()
+        except ResourceHandlingFailed as e:
+            logging.error(
+                "Unable to create ClusterRole for training-operator: %s", str(e)
+            )
+        else:
+            logging.info("Successfully created ClusterRole for training-operator")
+            self.unit.status = ActiveStatus()
```
jnsgruk, line 109: As above :)

src/charm.py
```diff
@@ -80,10 +82,52 @@ def _on_config_changed(self, _):
...
+            K8sResources.create_cluster_role_binding()
+        except ResourceHandlingFailed as e:
+            logging.error(
+                "Unable to create ClusterRoleBinding for training-operator: %s", str(e)
+            )
+        else:
+            logging.info(
+                "Successfully created ClusterRoleBinding for training-operator"
+            )
+            self.unit.status = ActiveStatus()
```
jnsgruk, line 125: As above :)

src/charm.py
```diff
@@ -80,10 +82,52 @@ def _on_config_changed(self, _):
...
 
+    def _create_service_account(self):
+        """Create a ServiceAccount for training-operator."""
+        self.unit.status = MaintenanceStatus("Creating ServiceAccount")
+        try:
+            K8sResources.create_service_account()
+        except ResourceHandlingFailed as e:
+            logging.error(
+                "Unable to create ServiceAccount for training-operator: %s", str(e)
+            )
```
jnsgruk, line 93: consider returning something from this function, like a `bool` or similar to indicate success or failure? Right now this essentially silently fails unless you're watching the logs. Same for other helpers.

src/charm.py
```diff
@@ -80,10 +82,52 @@ def _on_config_changed(self, _):
...
+                "Successfully created ClusterRoleBinding for training-operator"
+            )
+            self.unit.status = ActiveStatus()
 
     def _on_install(self, _):
-        """Placeholder for install event"""
-        pass
+        """Event handler for on-install events"""
+        self._create_service_account()
+        self._create_cluster_role_binding()
```
jnsgruk, line 130: If you do indeed make these helpers return something, you could do something like this:

```suggestion
        sa = self._create_service_account()
        crb = self._create_cluster_role_binding()
        
        if sa and crb:
            self.unit.status = ActiveStatus()
        else:
            self.unit.status = BlockedStatus("unable to create necessary resources, check logs")
```

src/kubernetes_resources.py
```diff
@@ -0,0 +1,189 @@
...
+
+        namespace = K8sResources.namespace()
+        name = K8sResources.name
+
+        # Populate metadata, role_ref, subjects, and body parameters
+        metadata = kubernetes.client.V1ObjectMeta(name=name, labels={"app": name})
+        role_ref = kubernetes.client.V1RoleRef(
+            api_group="rbac.authorization.k8s.io",
+            kind="ClusterRole",
+            name="training-operator",
```
jnsgruk, line 165:
```suggestion
            name=name,
```

src/kubernetes_resources.py
```diff
@@ -0,0 +1,189 @@
...
+        )
+        subject = kubernetes.client.V1Subject(
+            kind="ServiceAccount", name="training-operator", namespace=namespace
```
jnsgruk, line 168:
```suggestion
            kind="ServiceAccount", name=name, namespace=namespace
```

jnsgruk (changes requested):
Looking good, a few changes :)

src/kubernetes_resources.py
```diff
@@ -0,0 +1,189 @@
+#!/usr/bin/env python3
+
+import kubernetes
```
DomFleischmann, line 3: So I didin't feel entirely comfortable with this approach of creating K8s Resources and tried it out with lightkube.

[Here](https://github.com/canonical/training-operator/tree/lightkube) is a branch which shows how this would look like with lightkube and just plain manifests. I added some templating just to see if it works and it does. Of course this is just something I put together in an hour so feel free to change it and add more templating.

I think we should use this approach instead of the current one. It seems a lot more maintainable in the long run and more similar to what we will get with Juju Templates (I hope).

Pinging @jnsgruk @knkski and @ca-scribner so that they take a look.

src/kubernetes_resources.py
DnPlas: I guess none of this applies anymore. Good catch, though.

requirements.txt
```diff
@@ -1,3 +1,4 @@
 # Switch back to pypi package when ops is updated with `container.can_connect`
 # ops
 git+https://github.com/canonical/operator/#egg=ops
+kubernetes
```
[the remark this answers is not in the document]
DnPlas: +1 on pinning versions. Done in the new version of this PR. Resolving this.

src/charm.py
```diff
@@ -80,11 +85,45 @@ def _on_config_changed(self, _):
...
+        with open(os.path.join(self._src_dir, "auth_manifests.yaml")) as f:
+            for obj in codecs.load_all_yaml(f, context=context):
+                try:
+                    client.patch(
+                        type(obj), obj.metadata.name, obj, patch_type=PatchType.MERGE
+                    )
+                except ApiError as e:
+                    logging.error(f"Patching auth resources failed: {str(e)}")
+        logging.info("Auth resources successfully patched.")
+        return True
```
DomFleischmann, line 103: Unless I'm missing something, in the case of an ApiError you would log an error but still end up returning True. Seems like there is a `return False` statement missing.

src/charm.py
```diff
@@ -80,11 +85,45 @@ def _on_config_changed(self, _):
...
+        client = Client()
+
+        with open(os.path.join(self._src_dir, "crds_manifests.yaml")) as f:
+            for obj in codecs.load_all_yaml(f):
+                try:
+                    client.create(obj)
+                except ApiError as e:
+                    logging.error(f"Creating CRDs failed: {str(e)}")
+        logging.info("CRDs successfully created.")
+        return True
```
DomFleischmann, line 117: Same as above

src/charm.py
```diff
@@ -80,11 +85,45 @@ def _on_config_changed(self, _):
         else:
             self.unit.status = WaitingStatus("Waiting for Pebble in workload container")
 
+    def _patch_auth_resources(self):
+        """Returns True if existing auth resources for training-operator are patched correctly."""
+        self.unit.status = MaintenanceStatus("Patching auth configuration")
+        client = Client()
+        context = {"namespace": self._namespace}
+
+        with open(os.path.join(self._src_dir, "auth_manifests.yaml")) as f:
```
jnsgruk, line 94:
```suggestion
        with open(Path(self._src_dir) / "auth_manifests.yaml") as f:
```

Just a nit, but I prefer using `pathlib` where possible :)

src/charm.py
```diff
@@ -80,11 +85,45 @@ def _on_config_changed(self, _):
         else:
             self.unit.status = WaitingStatus("Waiting for Pebble in workload container")
 
+    def _patch_auth_resources(self):
```
jnsgruk, line 88:
```suggestion
    def _patch_auth_resources(self) -> bool:
```

src/auth_manifests.yaml
```diff
@@ -0,0 +1,71 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    app: training-operator
+  name: {{ namespace }}-training-operator
```
DomFleischmann, line 7: The `training-operator` part could also change depending on the application name with which it gets deployed. I would recommend templating the application name also. Something like:
```suggestion
    app: {{ app-name }}
  name: {{ namespace }}-{{ app-name }}
```

requirements.txt
```diff
@@ -1,3 +1,5 @@
 # Switch back to pypi package when ops is updated with `container.can_connect`
 # ops
 git+https://github.com/canonical/operator/#egg=ops
+lightkube<=0.8.0
+jinja2<3.1
```
DomFleischmann, line 5: Is this dependency actually required? I think I was able to template just with the `lightkube` dependency.

src/charm.py
jnsgruk, line 103: Do you always want to return `True`, even if there are some resources that have not successfully applied? 

You could consider a function variable named `success` or similar that starts `True`, but gets flipped in the exception handler above, then return `success`?

src/charm.py
jnsgruk, line 117: As above, this function always returns `True`...

src/charm.py
```diff
@@ -80,11 +85,45 @@ def _on_config_changed(self, _):
...
+                    logging.error(f"Patching auth resources failed: {str(e)}")
+        logging.info("Auth resources successfully patched.")
+        return True
+
+    def _create_crds(self):
+        """Returns True if CRDs for training-operator are created correctly."""
+        self.unit.status = MaintenanceStatus("Creating CRDs")
+        client = Client()
+
+        with open(os.path.join(self._src_dir, "crds_manifests.yaml")) as f:
```
jnsgruk, line 110:
```suggestion
        with open(Path(self._src_dir) / "crds_manifests.yaml") as f:
```

src/charm.py
```diff
@@ -80,11 +85,45 @@ def _on_config_changed(self, _):
...
     def _on_install(self, _):
-        """Placeholder for install event"""
-        pass
+        """Event handler for on-install events."""
+        auth = self._patch_auth_resources()
+        crds = self._create_crds()
 
+        # Check resources were applied/created correctly
+        if auth and crds:
+            self.unit.status = ActiveStatus()
```
jnsgruk, line 126: What if these have not successfully applied? What state do you want to leave the charm in?

Once the above methods are changed, one of these conditions could be `False`, which I think would leave the charm in `MaintenanceStatus`. I'm not actually sure what the correct status is here; you could put the charm into `BlockedStatus`, and then provide an action that tries to re-apply the manifests? 

Probably worth some discussion - in some cases (like in the K8s service patcher library), I just log and move on, because the Service patching doesn't *stop* the application from running. In the case that something is required for an app to even function, Blocked seems appropriate, but in the case that it's not critical, ActiveStatus with a message might be a good idea?

src/kubernetes_resources.py
ca-scribner: I'm pretty encouraged as well.  I'm interested in playing with their templating too as that could be very nice

src/auth_manifests.yaml
DnPlas: That's true, good catch! I'll change it (:

src/charm.py
DnPlas: Agree on the missing `return False`. Will work on it.

src/charm.py
knkski: It looks like the point of returning `bool` here is that we can skip over errors, and return a flag that says "something failed, but we patched as much as we could". Given that to a sysadmin there's no real difference between that and "i'm raising an Exception at the first thing that failed" in terms of the error requiring manual intervention, I think we should go with something like this:

```python
def _patch_auth_resources(self):
    ...
    with open('auth-stuff.yaml') as f:
        for obj in codes.load_all_yaml(f):
            client.patch(...)

def _on_install(self, event):
    try:
        self.unit.status = MaintenanceStatus("Patching auth configuration")
        self._patch_auth_resources()
        self.unit.status = MaintenanceStatus("Creating CRDs")
        self._create_crds()
        ... etc ...
        self.unit.status = ActiveStatus()
    except ApiError as err:
        logging.error("Applying resources failed.")
        logging.error(traceback.format_exc())
        self.unit.status = BlockedStatus(str(err))
```

That also has the advantage of only setting a status in the event handler, which makes it much easier overall to discover what states a charm can be in.

In a broader sense, I think any function that returns a `bool` should probably be named `is_*` or `has_*` and not do actual work. For better or worse, Python uses exceptions for control flow handling, we probably don't want to fight Python on that.

Also, I wonder if we can make our lives simpler by relying on the Kubernetes API server to guarantee atomicity here, so that we don't have to even worry about "most things worked, but not everything". Looking around, I see [this SO question](https://serverfault.com/q/1047438), which says this:

> All API requests are atomic; an operation either completes entirely or not at all

If we can shove the patching and CRD creation into a single API request (or at least all of the patching into one request, and CRD creation into another), that link makes it seem like we won't have to worry about partial success here. Is that something that's possible with `lightkube`?

src/auth_manifests.yaml
DnPlas: Updated in latest version. Resolving this.

src/charm.py
DnPlas: Updated it in latest PR version. Resolving this.

src/charm.py
DnPlas: Same as above.

src/charm.py
DnPlas: Updated it in latest PR version. Resolving this.

src/charm.py
DnPlas: The case when CRDs already exist on the cluster and no further action is needed is covered in `_crds_creation`. I think that one is not critical, so returning `True` there and setting `ActiveStatus` later in the event handler should be okay. 
For cases when resources were not created/patched correctly, +1 on using `BlockedStatus`. By definition it is a status that will inform users when something went wrong with config, and manual intervention is needed, though I am not sure about the manual intervention part, shall we just provide the exception and let users sort out the rest?

jnsgruk: Seems like a particularly interesting course to pursue in the context of how most of the KF charms work broadly outside of Juju's control.

DnPlas: Perhaps we don't need an extra `validate_resource` action as we are already iterating over the resources within `_create_crds()` and `_patch_auth_resources()`, and when an exception occurs, we know exactly what resource is having the issue as well as the failure reason, and other interesting and useful info. 

Regarding the action users should do in the event of failure, I like this idea here:
> ...checks if a corresponding resource exists on the cluster, and dumps the manifest if not?

That being said, we can set `BlockedStatus`, and dump the exception msg and manifest of the missing resource (we can even save the yaml file in a tmp dir) with instructions on how to create/patch the resource after the issue that caused the exception is resolved.

That, or a mechanism by which users can tell juju they've resolved the issue and let it re-try to patch/create resources again, without needing humans to manage any manifest files. Though for this I wouldn't know how to do it, perhaps in the config-changed event handler.

knkski: I agree with @DnPlas that for now, we should set `BlockedStatus`, and log the resource that failed to create along with the traceback. That will be enough for an admin to come in and know what's wrong. I like where @jnsgruk and @ca-scribner are going with this in general, but I think a lot of that work is something that Juju should negotiate with Kubernetes, and not something a charm author should have to write themselves.

For example, I don't think every K8s charm should be copy/pasting a `validate_resource` action, that's the sort of thing Juju itself should provide. Particularly with Juju Templates, we could tell Juju "create these resources", and if any of them failed to create, Juju would automatically make that apparent in `juju status`. We could then have some sort of `juju show-failed-resources` command to make it easily visible what failed to create and why.

src/charm.py
```diff
@@ -80,11 +86,50 @@ def _on_config_changed(self, _):
...
+        logging.info("Auth resources successfully patched.")
+        return True
+
+    def _create_crds(self) -> bool:
+        """Returns True if CRDs are created correctly, False otherwise."""
+        self.unit.status = MaintenanceStatus("Creating CRDs")
+        client = Client()
+
+        with open(Path(self._src_dir) / "crds_manifests.yaml") as f:
+            for obj in codecs.load_all_yaml(f):
```
knkski, line 113: It really seems like we should be propagating the errors and letting the event handler catch them, for something like this:

```python
try:
    client.create(obj)
except ApiError as e:
    if e.status.reason == 'AlreadyExists:
        # ignore
    else:
        raise
```
and
```python
  try:
      ...
  except ApiError as err:
      logging.error("Applying resources failed.")
      logging.error(traceback.format_exc())
      self.unit.status = BlockedStatus(str(err))
```

Returning a `bool` here doesn't tell the caller anything about why it failed, which means the `_on_install` handler can't set a useful `BlockedStatus` message.

DnPlas: As of now, `_on_install` doesn't handle the case where either `_create_crds()` or `_patch_auth` return `False`, but that's something I'll add once we decide how to correctly tell users what to do in the event of failure. I am thinking on something like:
`_create_crds():`

```python
  try:
  ...
  except ApiError as err:
      logging.error("<specific resource> failed.") # Let users know exactly which resource it tried to patch/create
      logging.error(err.<data>) # Exception data
      return False
```
`_on_install():`

```python
if True:
  self.unit.status = ActiveStatus()
else:
  self.unit.status = BlockedStatus("Resources failed, please do <action>"))
```

As someone commented previously, logging is okay in helper methods, but setting the status should happen in the event handler. Does this make sense?

src/charm.py
DnPlas: I agree with @knkski that Juju should provide this kind of functionality. I think k8s native charms will be benefited by something like `juju validate-resources` and/or `juju show-failed-resources`. I'd like to see something like that, though extending juju's capabilities are out of the scope of this PR :) 
For now, I'm going to submit a new version of this PR letting the helper methods show all useful data, and the event handler set the status. As always, I'd appreciate your comments.

DnPlas: One extra thing. I would like to know your thoughts on using `juju resolved`. If we provide enough info about what is causing the issue, can we go with something like `BlockedStatus("Resolve <issues> and run juju resolved afterwards)`?

src/charm.py
jnsgruk: Agree on the latter point, but I think what @knkski is saying is, you could:

```python
try:
    helper_methods
except ApiError as e:
    logging.error("something")
    self.unit.status = BlockedStatus("helllllllp")
else:
    self.unit.status = ActiveStatus()
```

DnPlas: Oh, I get it. Then +1.

src/charm.py
```diff
@@ -21,8 +26,9 @@ def __init__(self, *args):
         self._name = self.model.app.name
         self._namespace = self.model.name
         self._manager_service = "manager"
+        self._src_dir = os.path.dirname(__file__)
```
jnsgruk, line 29:
```suggestion
        self._src_dir = Path(__file__).parent
```

We're using Pathlib elsewhere, so... :)

src/charm.py
```diff
@@ -80,11 +86,68 @@ def _on_config_changed(self, _):
         else:
             self.unit.status = WaitingStatus("Waiting for Pebble in workload container")
 
+    def _patch_auth_resources(self) -> None:
+        """Patch the Kubernetes auth resources created by Juju.
 
-    def _on_install(self, _):
-        """Placeholder for install event"""
-        pass
+        Rases:
```
jnsgruk, line 92:
```suggestion
        Raises:
```

jnsgruk (approved):
Minor nits, looks good :)

src/charm.py
DnPlas: Ah shoot, that again haha! I forgot :D Will change it.

src/charm.py
DnPlas: Done. Resolving this.

src/charm.py
```diff
@@ -80,11 +86,68 @@ def _on_config_changed(self, _):
...
+                        raise
 
+    def _on_install(self, _):
+        """Event handler for on-install events."""
+        try:
+            self._patch_auth_resources()
+            self._create_crds()
+        except ApiError as e:
+            logging.error(
+                f"Some Kubernetes resources failed to be created/patched: {str(e)}"
```
knkski, line 144: Can we use `traceback.format_exc()` here instead of `str(e)`? That will log the full traceback, including if there's multiple exceptions involved

src/charm.py
```diff
@@ -80,11 +86,68 @@ def _on_config_changed(self, _):
...
+                try:
+                    client.create(obj)
+                    logging.info(f"{obj.metadata.name} CRD successfully created.")
+                except ApiError as e:
+                    if e.status.reason == "AlreadyExists":
+                        logging.info(
+                            f"{obj.metadata.name} CRD already present. It will be used by the operator."
+                        )
+                    else:
+                        logging.error(
```
knkski, line 132: I think we can skip this logging statement if we use `traceback.format_exc()` as in my other comment. The traceback includes the description of the object, including its name and why it failed. As an example where I tried to set an invalid label, I got this:

```python
... rest of traceback ...
lightkube.core.exceptions.ApiError: CustomResourceDefinition.apiextensions.k8s.io "profiles.kubeflow.org" is invalid: [metadata.labels: Invalid value: "\x00": name part must consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyName',  or 'my.name',  or '123-abc', regex used for validation is '([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]'), metadata.labels: Invalid value: "!": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?'), spec.versions[0].schema.openAPIV3Schema: Required value: schemas are required, spec.versions[1].schema.openAPIV3Schema: Required value: schemas are required]
```

I think that's enough information that we don't have to log it here as well.

src/charm.py
```diff
@@ -80,11 +86,68 @@ def _on_config_changed(self, _):
...
+        """Event handler for on-install events."""
+        try:
+            self._patch_auth_resources()
+            self._create_crds()
+        except ApiError as e:
+            logging.error(
+                f"Some Kubernetes resources failed to be created/patched: {str(e)}"
+            )
+            self.unit.status = BlockedStatus(
+                f"Creating/patching resources failed with code {str(e.status.code)}. Please resolve the issue(s) and run juju resolved to retry."
```
knkski, line 147: ping @johnsca on this. Can we blanket recommend `juju resolved` here?

src/charm.py
DnPlas: Done :)

src/charm.py
DomFleischmann: I feel like logging the failed resource and asking the user to run juju resolved afterwards is the right approach. Please add these changes and let's get this PR merged.

src/auth_manifests.yaml
```diff
@@ -0,0 +1,64 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    app: {{ app_name }}
+  name: {{ app_name }}
```
ca-scribner, line 7: I think for ClusterRoles we should add a bit more context to the naming since they're globally scoped (I realise that you can't atm deploy two kubeflows to the same k8s cloud, but maybe some day ... :)  ).  What about this?
```suggestion
  name: {{ namespace }}:{{ app_name }}
```

Could also do "-".  I see both ":" and "-" being used by different applications in `kubectl get clusterrolebinding`

src/auth_manifests.yaml
```diff
@@ -0,0 +1,64 @@
...
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  labels:
+    app: {{ app_name }}
+  name: {{ app_name }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: {{ app_name }}
```
ca-scribner, line 60: Same as above for ClusterRole

PR #6, opened 2021-10-15
Adds integration tests, testing dependencies and tox.ini for integration testing

DnPlas:
This PR introduces the following changes:
* tests/integration: adds integration tests
* Adds tox.ini and test-requirements.txt
* Adds examples/ with simple training jobs manifests (excluding mxnet)

> NOTE: I understand that multiple asserts in a single test is not desirable, please let me know if you've an alternative for `test_can_i_create_training_jobs`.

### DIFF
--- a/tests/integration/test_charm.py
+++ b/tests/integration/test_charm.py
@@ -106,7 +106,8 @@ def test_create_training_jobs(ops_test: OpsTest, example: str):
         assert job_status in [
             "Running",
             "Succeeded",
-        ], f"{job_object.metadata.name} was not running or did not succeed"
+        ], f"{job_object.metadata.name} was not running or did not succeed" \
+        f" (status == {job_status})"
 
     assert_get_job()
     assert_job_status_running_success()
--- a/tests/integration/test_charm.py
+++ b/tests/integration/test_charm.py
@@ -1,0 +1,112 @@
+import glob
+from pathlib import Path
+
+import yaml
+import pytest
+import tenacity
+import lightkube.generic_resource
+
+from lightkube import codecs, Client
+from pytest_operator.plugin import OpsTest
+from tenacity import retry, wait_exponential, stop_after_delay
+
+METADATA = yaml.safe_load(Path("./metadata.yaml").read_text())
+APP_NAME = "training-operator"
+
+@pytest.mark.abort_on_fail
+async def test_build_and_deploy(ops_test: OpsTest):
+    """Build the charm and deploy it with trust=True.
+
+    Assert on the unit status.
+    """
+    charm_under_test = await ops_test.build_charm(".")
+    image_path = METADATA["resources"]["training-operator-image"]["upstream-source"]
+    resources = {"training-operator-image": image_path}
+
+    await ops_test.model.deploy(
+        charm_under_test, resources=resources, application_name=APP_NAME, trust=True
+    )
+    await ops_test.model.wait_for_idle(
+        apps=[APP_NAME], status="active", raise_on_blocked=True, timeout=60 * 10
+    )
+    assert ops_test.model.applications[APP_NAME].units[0].workload_status == "active"
+
+def lightkube_create_global_resources() -> dict:
+    """Returns a dict with GenericNamespacedResource as value for each CRD key."""
+    crds_kinds = [
+        crd["spec"]["names"]
+        for crd in yaml.safe_load_all(Path("./src/crds_manifests.yaml").read_text())
+    ]
+    jobs_classes = {}
+    for kind in crds_kinds:
+        job_class = lightkube.generic_resource.create_namespaced_resource(
+            group="kubeflow.org", version="v1", kind=kind["kind"], plural=kind["plural"]
+        )
+        jobs_classes[kind["kind"]] = job_class
+    return jobs_classes
+
+# TODO: Kubeflow upstream MXNet examples use GPU.
+# Not testing MXNetjobs until we have a CPU mxjob example.
+JOBS_CLASSES = lightkube_create_global_resources()
+
+@pytest.mark.parametrize("example", glob.glob("examples/*.yaml"))
+def test_create_training_jobs(ops_test: OpsTest, example: str):
+    """Validates that a training job can be created and is running.
+
+    Asserts on the *Job status.
+    """
+    namespace = ops_test.model_name
+    lightkube_client = lightkube.Client()
+
+    # Set up for creating an object of kind *Job
+    job_yaml = yaml.safe_load(Path(example).read_text())
+    job_object = lightkube.codecs.load_all_yaml(yaml.dump(job_yaml))[0]
+    job_class = JOBS_CLASSES[job_object.kind]
+
+    # Create *Job and check if it exists where expected
+    lightkube_client.create(job_object, namespace=namespace)
+
+    # Allow the resource to be created
+    @tenacity.retry(
+        wait=tenacity.wait_exponential(multiplier=1, min=1, max=15),
+        stop=tenacity.stop_after_delay(30),
+        reraise=True,
+    )
+    def assert_get_job():
+        """Asserts on the job.
+
+        Retries multiple times using tenacity to allow time for the training job
+        to be created.
+        """
+        job = lightkube_client.get(
+            job_class, name=job_object.metadata.name, namespace=namespace
+        )
+
+        assert job is not None, f"{job_object.metadata.name} does not exist"
+
+    # Wait for the *Job to have a status
+    # TODO: change this workaround after
+    # we have an implementation in lightkube
+    @tenacity.retry(
+        wait=tenacity.wait_exponential(multiplier=2, min=1, max=120),
+        stop=tenacity.stop_after_attempt(3),
+        reraise=True
+    )
+    def assert_job_status_running_success():
+        """Asserts on the job status.
+
+        Retries multiple times using tenacity to allow time for the training job
+        to change its status from None -> Created -> Running/Succeeded.
+        """
+        job_status = lightkube_client.get(
+            job_class.Status, name=job_object.metadata.name, namespace=namespace
+        ).status["conditions"][-1]["type"]
+
+        # Check wether the last status of *Job is Running/Success
+        assert job_status in [
+            "Running",
+            "Succeeded",
+        ], f"{job_object.metadata.name} was not running or did not succeed"
+
+    assert_get_job()
+    assert_job_status_running_success()
--- a/examples/pytorchjob.yaml
+++ b/examples/pytorchjob.yaml
@@ -1,0 +1,32 @@
+apiVersion: "kubeflow.org/v1"
+kind: PyTorchJob
+metadata:
+  name: pytorch-simple
+spec:
+  pytorchReplicaSpecs:
+    Master:
+      replicas: 1
+      restartPolicy: OnFailure
+      template:
+        spec:
+          containers:
+            - name: pytorch
+              image: docker.io/kubeflowkatib/pytorch-mnist:v1beta1-45c5727
+              imagePullPolicy: Always
+              command:
+                - "python3"
+                - "/opt/pytorch-mnist/mnist.py"
+                - "--epochs=1"
+    Worker:
+      replicas: 1
+      restartPolicy: OnFailure
+      template:
+        spec:
+          containers:
+            - name: pytorch
+              image: docker.io/kubeflowkatib/pytorch-mnist:v1beta1-45c5727
+              imagePullPolicy: Always
+              command:
+                - "python3"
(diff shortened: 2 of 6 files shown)


tests/integration/test_charm.py
```diff
@@ -0,0 +1,80 @@
...
+@pytest.mark.abort_on_fail
+async def test_can_i_create_training_jobs(ops_test: OpsTest):
+    """Validate ability to interact with training jobs with current auth config.
+
+    Asserts whether training-operator can create training jobs with current auth config.
+    """
+    training_jobs = ("tfjobs", "pytorchjobs", "mxjobs", "xgboostjobs")
+    for tjob in training_jobs:
+        kubectl_get = (
+            "microk8s",
```
DomFleischmann, line 68: I would prefer if these test could run on other k8s also.

You can check this [PR](https://github.com/canonical/istio-operators/pull/34) from @johnsca for more info.

DnPlas:
Hi team. Thanks for the reviews, I've updated the PR. I think this new version is closer to what the general consensus is. 

* I left just one test case that asserts on the ability to create the `*Job`, and also checks its status. Please note that I am not checking pod(s) statuses directly, but through the Kubeflow resource itself, as each training job has its own spec and it would be different from job to job (for instance TFJobs have `PS` pods, but PyTorchJobs don't). This test case checks if the job exists, if it was `Created` and if its final state is either `Running` or `Succeeded`. Also note the program running in the container can fail (e.g. missing runtime deps for the training algo) and the pod status could be `Error`, but that should not be a problem for us.
* I used tenacity inspired by @ca-scribner's [integration testing](https://github.com/canonical/kubeflow-profiles-operator/blob/9d6e97515ffa7af227aa7bce4c38b66d1f4dcf3c/tests/integration/test_charm.py#L166).
* As @knkski suggested, I am not explicitly listing the example yaml files, but instead taking whatever is inside `examples`, which will help us to scale and change our examples without worrying about the test. I think I left everything pretty generic, but if you catch something, please let me know.
* I left one setup method for creating `GenericNamespacedResources`, which will allow us to get these `*Jobs` using lightkube. I don't want to include that as part of the test case because it is not something that belongs to the test itself. 

I believe this new version resolves all open conversations, but feel free to re-open them if necessary.

DnPlas:
> > @DnPlas: Regarding
> > > NOTE: I understand that multiple asserts in a single test is not desirable, please let me know if you've an alternative for test_can_i_create_training_jobs.
> > 
> > 
> > I would disagree that multiple asserts in a single test is a bad thing. A good test does something, then asserts as much as is reasonable about the current state of things. See here for a specific example:
> > https://github.com/canonical/minio-operator/blob/51a184e/tests/unit/test_charm.py#L112-L120
> > We assert that the charm is in ActiveStatus, then assert that the charm also sent some known data over the relation, and also assert that the secret key sent over was 30 characters in length (since it's randomly generated, we can't assert equality).
> 
> Completely agree with @knkski here, tests should be separated by the specific case they are testing, the more assertions there are for ensuring the test is successful, the better. On the other hand, when there are multiple assert calls in one case it might be confusing which assertion has failed, this is where I'm in favor of adding assert messages where they make sense.

I meant the latter, multiple asserts in the same test case can be confusing. But thanks for the feedback.

tests/integration/test_charm.py
```diff
@@ -0,0 +1,112 @@
...
+def test_create_training_jobs(ops_test: OpsTest):
+    """Validates that a training job can be created and is running.
+
+    Asserts on the *Job status.
+    """
+    namespace = ops_test.model_name
+    lightkube_client = lightkube.Client()
+    jobs_classes = lightkube_create_global_resources()
+
+    for p in glob.glob("examples/*.yaml"):
```
knkski, line 62: I liked @ca-scribner's idea to use pytest parameterization. Instead of the for loop here, can we decorate this function like this?

    @pytest.mark.parametrize("example", glob.glob("examples/*.yaml"))

IMO that's the best of both worlds.

tests/integration/test_charm.py
```diff
@@ -0,0 +1,92 @@
...
+    await ops_test.model.deploy(
+        charm_under_test, resources=resources, application_name=APP_NAME, trust=True
+    )
+    await ops_test.model.wait_for_idle(
+        apps=[APP_NAME], status="active", raise_on_blocked=True, timeout=60 * 10
+    )
+    assert ops_test.model.applications[APP_NAME].units[0].workload_status == "active"
+
+
+@pytest.mark.abort_on_fail
```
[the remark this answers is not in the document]
ca-scribner: yeah I think the idea of that since pytest typically runs tests independently, the annotation lets you say "well nothing else can pass if this fails so I should abort if this fails" about a test(s).  It could be used on anything, but really it is just used on tests that you want to exit after if they fail (so deployment is a good one, but if you've got 3 tests that all work independently then probably you want want it as you want to know if they each passed/failed every time)

tests/integration/test_charm.py
```diff
@@ -0,0 +1,112 @@
...
+
+    for p in glob.glob("examples/*.yaml"):
+        # Set up for creating an object of kind *Job
+        job_yaml = yaml.safe_load(Path(p).read_text())
+        job_object = lightkube.codecs.load_all_yaml(yaml.dump(job_yaml))[0]
+        job_class = jobs_classes[job_object.kind]
+
+        # Create *Job and check if it exists where expected
+        lightkube_client.create(job_object, namespace=namespace)
+        assert (
```
ca-scribner, line 70: Does lightkube's `.create()` hold the thread until the creation actually occurs?  If not, there could be a race condition here that a small wait or a retry would be good for.

tests/integration/test_charm.py
```diff
@@ -0,0 +1,112 @@
...
+            lightkube_client.get(
+                job_class, name=job_object.metadata.name, namespace=namespace
+            )
+            is not None
+        ), f"{job_object.metadata.name} does not exist"
+
+        # Wait for the *Job to have a status
+        # TODO: change this workaround after
+        # we have an implementation in lightkube
+        try:
```
ca-scribner, line 80: I think this try/except block can be omitted.  `retry` handles this for you.  You can move the `get_job_status` outside this test to a helper (or keep it here) then just call it directly without the try/except.

tests/integration/test_charm.py
```diff
@@ -0,0 +1,112 @@
...
+            job_status = get_job_status()
+
+        # Check whether the *Job was created
+        assert (
+            job_status[0]["type"] == "Created"
+        ), f"Something went wrong creating {job_object.metadata.name}"
+
+        # Check wether the *Job is in a Running/Success state
+        assert (
+            job_status[-1]["type"] == "Running" or job_status[-1]["type"] == "Succeeded"
```
ca-scribner, line 111: If there was a short delay (a few seconds) between job creation and job running/succeeded, what would you like this test to do?  As is, it will fail, but I think you'd want it to succeed (a few second delay isn't a sign of something broken, could just be a random computer thing that resolves itself...).  Also, are you confident that any job entering the `Running` state confirms the operator is working correctly?  If yes then this works, but if no then I'd say just test for "Succeeded" since I think we're using tests cases that we expect to succeed.

I think this assertion would do better with the same `retry` pattern as above.  For some acceptable period (something long enough you know if it takes that long things have gone wrong) check for the acceptable statuses (`succeeded` and maybe `running`) and if you don't see them retry.  This will handle any race conditions or pauses, and if you're looking for `succeeded` rather than `running` it also waits until the job actually finishes.  

You can also add something that exits the checks if it hits a known failure status (eg: `if status == 'error', dont retry and instead fail`)

tests/integration/test_charm.py
```diff
@@ -0,0 +1,92 @@
...
+    # TODO: Kubeflow upstream MXNet examples use GPU.
+    # Not testing MXNetjobs can be created correctly
+    # until we have a CPU mxjob example.
+    examples = ("pytorchjob.yaml", "tfjob.yaml", "xgboostjob.yaml")
+    for example in examples:
+        assert create_training_job(example, namespace=ops_test.model_name) == True
+
+
+@pytest.mark.abort_on_fail
+async def test_training_jobs_status(ops_test: OpsTest):
```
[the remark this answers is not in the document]
DnPlas: This does not apply anymore, resolving conversation.

tests/integration/test_charm.py
DnPlas: Updated it, thanks @ca-scribner

tests/integration/test_charm.py
ca-scribner: yeah if you did the `tenacity.retry` I think you can do something like a try every 5s, fail after 1m (or whatever values you come up with).  Anything like that that tries for more than 10-15s is probably covering all the possible cases

DnPlas: +1

tests/integration/test_charm.py
ca-scribner: Re status: I might have misunderstood the jobs.  I'd think they all have convergence or epoch limits that would end them but I might be wrong.  The pytorchjob at least looks like it will end.  I've never used xgboost though so not sure about that one.  Testing success sounds better than just running to me, but we can start with checking for either and see if it causes trouble.  

My hesitance comes from not knowing if the operator can still be broken after we see a `running` state (like if some data storage or I don't know what isn't wired properly), but let's just try it as is and see what happens :)

DnPlas: Nvm @ca-scribner , you are right, eventually all jobs will have `Succeeded`. 
We may want to keep `Running` because:
1) `Succeeded` depends on so much things that an arbitrary wait may not allow to happen. Having different examples and setting a wait that covers all cases could be problematic. For reference, [Conditions].(https://www.kubeflow.org/docs/components/training/tftraining/#conditions)
2) "`TFJobRunning` means all sub-resources (e.g. services/pods) of this TFJob have been successfully scheduled and launched and the job is running." which sounds enough to me.

ca-scribner: sounds good to me too then, thanks!

DomFleischmann (approved):
This now LGTM

tests/integration/test_charm.py
```diff
@@ -0,0 +1,118 @@
...
+
+        # Stop retrying if *Job status is Failed
+        if job_status == "Failed":
+            return "Failed"
+
+        # Check wether the last status of *Job is Running/Success
+        assert job_status["type"] in [
+            "Running",
+            "Succeeded",
+        ], f"{job_object.metadata.name} is not running or has not succeeded"
```
ca-scribner, line 118: just a small problem: the test doesn't use `assert_get_job` or `assert_job_status_running_success` :)  But otherwise this looks great!  This is not tested but I think almost what is needed
```suggestion
        ], f"{job_object.metadata.name} is not running or has not succeeded"
        
        assert_get_job()
        
        status = assert_job_status_running_success()
        if status == "Failed":
          raise (some sort of appropriate error here)
```

tests/integration/test_charm.py
```diff
@@ -0,0 +1,115 @@
...
+
+        Retries multiple times using tenacity to allow time for the training job
+        to change its status from None -> Created -> Running/Succeeded.
+        """
+        job_status = lightkube_client.get(
+            job_class.Status, name=job_object.metadata.name, namespace=namespace
+        ).status["conditions"][-1]["type"]
+        # Stop retrying if *Job status is Failed
+        if job_status == "Failed":
+            raise AssertionError("Failed != Running/Succeeded")
```
ca-scribner, line 106: This no longer does anything since the `retry_if_not_exception_message` was removed
```suggestion
```

DnPlas: +1

tests/integration/test_charm.py
DnPlas: Updated, resolving this conversation.

tests/integration/test_charm.py
```diff
@@ -0,0 +1,112 @@
...
+        """
+        job_status = lightkube_client.get(
+            job_class.Status, name=job_object.metadata.name, namespace=namespace
+        ).status["conditions"][-1]["type"]
+
+        # Check wether the last status of *Job is Running/Success
+        assert job_status in [
+            "Running",
+            "Succeeded",
+        ], f"{job_object.metadata.name} was not running or did not succeed"
```
ca-scribner, line 109: help us debug things in future :)
```suggestion
        ], f"{job_object.metadata.name} was not running or did not succeed" \
        f" (status == {job_status})"
```

tests/integration/test_charm.py
DnPlas: Updated it. Resolving conversation.

ca-scribner (approved):

tests/integration/test_charm.py
DnPlas: Resolving conversation.

PR #8, opened 2021-10-22
tests/unit: adds unit testing for training-operator

DnPlas:
This PR adds unit testing for the training-operator. These commits introduce the following changes:
* tests/unit: adds unit testing for training-operator
    * Test cases are divided into two groups: handlers to exercise event handlers, and helpers. Please note these test cases rely on what's defined in [PR#7](https://github.com/canonical/training-operator/pull/7)
    * Mocks based on @jnsgruk's example for [charm-kubernetes-dashboard](https://github.com/jnsgruk/charm-kubernetes-dashboard/blob/8b2a7320dcc51f8a042a3eaddae583f6825eedea/tests/unit/test_charm.py)
* tox.ini: adds unit test block

Feedback is greatly appreciated

### DIFF
--- a/tests/unit/test_charm.py
+++ b/tests/unit/test_charm.py
@@ -1,0 +1,162 @@
+from pathlib import Path
+import unittest
+from glob import glob
+from unittest.mock import MagicMock, Mock, patch
+import logging
+
+import lightkube
+from lightkube import codecs
+from lightkube.core.exceptions import ApiError
+from ops.model import ActiveStatus, BlockedStatus
+from ops.testing import Harness
+
+from charm import TrainingOperatorCharm
+
+logger = logging.getLogger(__name__)
+
+
+class _FakeResponse:
+    """Used to fake an httpx response during testing only."""
+
+    def __init__(self, code):
+        self.code = code
+        self.name = ""
+
+    def json(self):
+        reason = ""
+        if self.code == 409:
+            reason = "AlreadyExists"
+        return {
+            "apiVersion": 1,
+            "code": self.code,
+            "message": "broken",
+            "reason": reason,
+        }
+
+
+class _FakeApiError(ApiError):
+    """Used to simulate an ApiError during testing."""
+
+    def __init__(self, code=400):
+        super().__init__(response=_FakeResponse(code))
+
+
+@patch("lightkube.core.client.GenericSyncClient", Mock)
+class TestCharm(unittest.TestCase):
+    def setUp(self):
+        self.harness = Harness(TrainingOperatorCharm)
+        self.addCleanup(self.harness.cleanup)
+        self.harness.begin()
+
+    # Event handlers
+    @patch("charm.TrainingOperatorCharm._create_crds")
+    @patch("charm.TrainingOperatorCharm._create_resource")
+    def test_install_event(self, create_auth, create_crds):
+        self.harness.charm.on.install.emit()
+
+        # Ensure create_auth and create_crds
+        # are called once
+        create_auth.assert_called_once()
+        create_crds.assert_called_once()
+
+        # Check status is Active
+        self.assertEqual(self.harness.charm.unit.status, ActiveStatus())
+
+    @patch("charm.TrainingOperatorCharm._create_crds")
+    @patch("charm.TrainingOperatorCharm._create_resource")
+    @patch("charm.ApiError", _FakeApiError)
+    def test_blocked_on_appierror_on_install_event(self, create_auth, create_crds):
+        # Ensure the unit is in BlockedStatus
+        # on exception when creating auth resources
+        subtests = (
+            (
+                create_auth,
+                "Test BlockedStatus when ApiError is raised trying to create auth resources",
+            ),
+            (
+                create_crds,
+                "Test BlockedStatus when ApiError is raised trying to create CRDS",
+            ),
+        )
+        for create_type, subtest_description in subtests:
+            with self.subTest(msg=subtest_description):
+                create_type.side_effect = _FakeApiError()
+                try:
+                    self.harness.charm.on.install.emit()
+                except ApiError:
+                    self.assertEqual(
+                        self.harness.charm.unit.status,
+                        BlockedStatus(
+                            f"Creating/patching resources failed with code {create_type.side_effect.response.code}."
+                        ),
+                    )
+
+    @patch("charm.TrainingOperatorCharm._update_layer")
+    @patch("charm.TrainingOperatorCharm._patch_resource")
+    @patch("charm.ApiError", _FakeApiError)
+    def test_config_changed_event(self, patch, update):
+        self.harness.charm.on.config_changed.emit()
+
+        # Ensure _update_layer is called once and
+        # _patch_resource is called
+        update.assert_called_once()
+        patch.assert_called()
+
+        # Ensure the unit is in BlockedStatus
+        # on exception when patching auth resources
+        patch.side_effect = _FakeApiError()
+        try:
+            self.harness.charm.on.config_changed.emit()
+        except ApiError:
+            self.assertEqual(
+                self.harness.charm.unit.status,
+                BlockedStatus(
+                    f"Patching resources failed with code {patch.side_effect.response.code}."
+                ),
+            )
+
+    @patch("charm.TrainingOperatorCharm._update_layer")
+    def test_on_training_operator_pebble_ready(self, update):
+        self.harness.container_pebble_ready("training-operator")
+
+        # Check the layer gets created
+        self.assertIsNotNone(
+            self.harness.get_container_pebble_plan("training-operator")._services
+        )
+
+    # Helpers
+    @patch("charm.Client.create")
+    def test_create_resource(self, client: MagicMock):
+        subtests = (
+            (
+                "auth",
+                {"namespace": "unit-kubeflow", "app_name": "unit-training-operator"},
+                "auth_manifests.yaml",
+            ),
+            ("crds", {}, "crds_manifests.yaml"),
+        )
+        for resource_type, context, template in subtests:
+            with self.subTest(msg=f"Testing {resource_type} creation"):
+                self.harness.charm._create_resource(
+                    resource_type=resource_type, context=context
+                )
+
+                # Ensure client is called to create resources in src/*.yaml
+                with open(f"./src/{template}") as f:
+                    for resource in codecs.load_all_yaml(f, context):
+                        client.assert_any_call(resource)
+
+
+    @patch("charm.TrainingOperatorCharm._create_resource")
+    @patch("charm.ApiError", _FakeApiError)
+    def test_create_crds(self, create_resource):
+        self.harness.charm._create_crds()
+        create_resource.assert_called_once()
+
+        # Check ApiError is raised
+        create_resource.side_effect = _FakeApiError()
+        with self.assertRaises(ApiError):
+            self.harness.charm._create_crds()
+
+        # TODO: Ensure 'CRD already present' message is logged on
+        # ApiError for reason='AlreadyExists'
--- a/tox.ini
+++ b/tox.ini
@@ -11,7 +11,9 @@ passenv =
     KUBECONFIG
 
 [testenv:unit]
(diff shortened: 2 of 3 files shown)


tests/unit/test_charm.py
```diff
@@ -0,0 +1,145 @@
...
+        # on exception when creating auth resources
+        create_auth.side_effect = _FakeApiError()
+        self.harness.charm.on.install.emit()
+        self.assertEqual(self.harness.charm.unit.status, BlockedStatus(f"Creating/patching resources failed with code {create_auth.side_effect.response.code}."))
+
+        # Ensure the unit is in BlockedStatus
+        # on exception when creating CRDs
+        create_crd.side_effect = _FakeApiError()
+        self.harness.charm.on.install.emit()
+        self.assertEqual(self.harness.charm.unit.status, BlockedStatus(f"Creating/patching resources failed with code {create_crd.side_effect.response.code}."))
```
ca-scribner, line 59: This may just be a style thing, but this feels like three separate tests to me, one for default `install` behaviour and then two when errors aren't handled properly.  Those feel like things to independently test as they can independently break

ca-scribner: A concrete example of why I'd break this up too is that I think the third test is not actually testing the `create_crd` error catch.  I believe the `self.hardness.charm.on.install.emit()` that is intended to test `create_crd` is actually failing because `create_auth` is mocked and still raising errors (it gets checked before `create_crd`).  A way to prove this would be to put different error codes in the different side effects, but personally I'd break them up into separate tests

tests/unit/test_charm.py
```diff
@@ -0,0 +1,145 @@
...
+    def test_config_changed_event(self, patch, update):
+        self.harness.charm.on.config_changed.emit()
+        update.assert_called_once()
+
+        # Run each case of patch here?
+        patch.assert_called()
+
+        # Ensure the unit is in BlockedStatus
+        # on exception when patching auth resources
+        patch.side_effect = _FakeApiError()
```
ca-scribner, line 72: Same comment as above.  This feels like a separate test that can break independently of the first half

tests/unit/test_charm.py
```diff
@@ -0,0 +1,145 @@
...
+        self.harness.charm._context = {"namespace": "kubeflow", "app_name": "unit-training-operator"}
+        self.harness.charm._create_resource(resource_type="auth", context=self.harness.charm._context)
+        self.harness.charm._create_resource(resource_type="crds")
+
+        # Ensure all resources listed in src/*.yaml exist
+        resources = []
+        for manifest in glob("src/*.yaml"):
+            with open(manifest) as f:
+                resources.extend([r for r in codecs.load_all_yaml(f, self.harness.charm._context)])
+        for resource in resources:
```
ca-scribner, line 96: I don't know if there's an equivalent in UnitTest, but this seems like where I'd use a parametrized test in pytest, testing only a single resource call at a time.  Something like:

(syntax is probably wrong, just going by memory)
```python
@patch("charm.Client.create")
@pytest.mark.parametrize(
  "resource,context,yaml_files",
  [
    ("auth", {"namespace": "kubeflow, ...}, [yaml_file1, ...]),
    ("crds", {}, [yaml_file1, ...]),
  ]
)
def test_create_resource(resource, context, yaml_files, client: MagicMock):
  self.harness.charm._create_resource(resource_type=resource, context=context)
  resources = []
  for manifest in yaml_files:
  ...
```

(and then do the failure tests separately, again parametrized if you want to test both types)

ca-scribner:
My comments were mainly stylistic.  I personally like tests that are very granular and testing a single purpose (not necessarily a single thing, eg `install`, but a single outcome, eg did `install` work when I gave it inputs like X) and then parametrizing.  That way when something fails, it isn't that `test_install` failed but that `test_install_catches_bad_resource_input` failed.

ca-scribner:
Testing for failures here felt natural (when call `X` returns code `broken`, check that we set `BlockedStatus`), but I had a hard time thinking of what testing for success should look like.  I got the feeling that you did as well.  A lot of the success testing was more "did I call the things I should?  that is probably ok then", but I'm not sure what to do to make it better :)

DnPlas:
> Testing for failures here felt natural (when call `X` returns code `broken`, check that we set `BlockedStatus`), but I had a hard time thinking of what testing for success should look like. I got the feeling that you did as well. A lot of the success testing was more "did I call the things I should? that is probably ok then", but I'm not sure what to do to make it better :)

Absolutely. We have all these methods that are important and part of the main function for charms, but they are not as easy to test. I will explore more options, like TDD, as these charms evolve so we make sure we cover most cases and actually test them.

tests/unit/test_charm.py
DnPlas: We can so something similar in unittest with [subTest()](https://docs.python.org/3/library/unittest.html#unittest.TestCase.subTest). I will explore this option, I like it better.

tests/unit/test_charm.py
DnPlas: You are right, the last assertion is not right, definitely will change that. This may be also a good place to use subTest(), I'll experiment with it.

tests/unit/test_charm.py
addyess: this is correct syntax.  I prefer this too and use it quite frequently

tox.ini
```diff
@@ -10,6 +10,9 @@ deps =
 passenv =
     KUBECONFIG
 
+[testenv:unit]
+commands = pytest -v {toxinidir}/tests/unit
```
addyess, line 14: consider reporting coverage at this point.  If you've got unit tests you can get the coverage with `pytest-cov` and using `--cov-report=term-missing` as an argument to get nice reports of what isn't unit tested -- even block a PR in the event the coverage isn't high enought

addyess:
maybe you could also add a github action to run the unit testing on PRs?  
See https://github.com/charmed-kubernetes/charm-kubernetes-master/blob/master/.github/workflows/main.yml#L5

tox.ini
ca-scribner: good suggestion.  I dont think we've done coverage on any of our charm repos yet

DnPlas:
> maybe you could also add a github action to run the unit testing on PRs? See https://github.com/charmed-kubernetes/charm-kubernetes-master/blob/master/.github/workflows/main.yml#L5

I rather keep PRs small, but sure that's the next thing in the queue. Thanks.

tests/unit/test_charm.py
```diff
@@ -0,0 +1,77 @@
+import logging
+import unittest
```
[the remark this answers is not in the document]
DnPlas: I agree with @DomFleischmann, unittest offers enough functionality to do these tests, though after talking a bit with @ca-scribner there could be interesting options we can explore with pytest. For now I'll leave unittest. Resolving this conversation.

tests/unit/test_charm.py
```diff
@@ -0,0 +1,77 @@
...
+class TestWithoutInitialHooks(unittest.TestCase):
+    def setUp(self):
+        self.harness = Harness(TrainingOperatorCharm)
+        self.addCleanup(self.harness.cleanup)
+        self.harness.begin()
+
+    # Tests related to Kubernetes resources
+    def test_kubernetes_auth_resources_created(self):
+        self.harness.charm.on.install.emit()
+        kubectl_get = ["microk8s.kubectl", "get",]
```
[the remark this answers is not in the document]
DnPlas: I updated the unit test. Thanks @jnsgruk for the examples. Resolving this conversation.

tests/unit/test_charm.py
DomFleischmann: Agree with @ca-scribner, This should be 3 different tests.

tests/unit/test_charm.py
DomFleischmann: +1

tox.ini
DomFleischmann: +1 I would like us to add coverage at some point.

tests/unit/test_charm.py
DnPlas: Yeah, maybe not three, but two: one testing on_install basic stuff, like `assert_called_once()`, and the other one testing `BlockedStatus` for each time a diff creation method fails. I decided to go with unittest.subTest() for the latter. Let me know what you think of this approach.

tox.ini
DnPlas: Added, thanks for the suggestion @addyess.

tests/unit/test_charm.py
```diff
@@ -0,0 +1,160 @@
...
+        # Ensure _update_layer is called once and
+        # _patch_resource is called
+        update.assert_called_once()
+        patch.assert_called()
+
+        # Ensure the unit is in BlockedStatus
+        # on exception when patching auth resources
+        patch.side_effect = _FakeApiError()
+        try:
+            self.harness.charm.on.config_changed.emit()
```
knkski, line 107: It looks like this is trying to test what happens on a failure. If so, we might want to use `assertRaises` here:

```python
with self.assertRaises(ApiError):
    self.harness.charm.on.config_changed.emit()
self.assertEqual(...)
```

tests/unit/test_charm.py
```diff
@@ -0,0 +1,160 @@
...
+    def test_create_resource(self, client: MagicMock):
+        subtests = (
+            (
+                "auth",
+                {"namespace": "unit-kubeflow", "app_name": "unit-training-operator"},
+                "auth_manifests.yaml",
+            ),
+            ("crds", {}, "crds_manifests.yaml"),
+        )
+        for resource_type, context, template in subtests:
```
knkski, line 136: I haven't used `unittest` over `pytest` in a while, but this section is IMO a point in favor of `pytest`. The `@pytest.mark.parametrize()` decorator is a lot nicer than this. I agree with the above conversation that we don't need to update this PR, just noting for future bikeshedding sessions :smile:

tox.ini
```diff
@@ -2,6 +2,7 @@
 skipsdist = True
 
 [testenv]
+basepython = python3
```
knkski, line 5: Are we actually running into an issue with this? I haven't had to use this in any other `tox.ini` file that I've written/used.

tox.ini
```diff
@@ -10,6 +11,11 @@ deps =
 passenv =
     KUBECONFIG
 
+[testenv:unit]
+deps =
+    {[testenv]deps}
+commands = pytest -v --cov=. --cov-report=term-missing {toxinidir}/tests/unit
```
knkski, line 17: The `--cov=. --cov-report=term-missing` bits seem like something we'd want to store as an artifact in Github Actions, maybe with `--cov-report html`. If you're up for it, it should be as simple as this:

```yaml
    - name: Upload coverage artifacts
      if: failure()
      uses: actions/upload-artifact@v2
      with:
        name: coverage
        path: <COV-FILE-NAME>
```

Otherwise we can kick that can down the road

tests/unit/test_charm.py
DnPlas: Yes, that was my first approach, unfortunately the code won't allow us to do that. I have this in my queue of improvements for `src/charm.py`.

tox.ini
DnPlas: I inherited this from [alertmanager](https://github.com/canonical/alertmanager-operator/blob/main/tox.ini#L16) and didn't pay much attention to it. We can remove it.

tests/unit/test_charm.py
DnPlas: I know :D @ca-scribner and I were having this conversation on Friday, we might want to move in the near future to pytest. Also in my queue of improvements.

tox.ini
DnPlas: @knkski Yes, agree that would be a nice addition, I really like the coverage report :ok_hand: . I'll add that feature in another PR as this one is for adding unit testing exclusively.

tox.ini
DnPlas: Removed. Closing this.

addyess (approved):
LGTM :shipit:

knkski (approved):
LGTM

PR #9, opened 2021-11-10
Enables integration and publish workflows, test coverage, and linting

DnPlas:
This PR enables integration and publish Github workflows, and other needed changes for it to work properly.
Changes introduced by the PR:
* tests/integration/: changes wait_exponential max time
* .github/workflows: adds integration and publish workflows
* Adds test coverage, linting, and runtime deps for both

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -1,0 +1,77 @@
+name: Tests
+
+on:
+  pull_request:
+  schedule:
+    - cron: '0 8 * * TUE'
+
+jobs:
+  lint:
+    name: Lint
+    runs-on: ubuntu-latest
+    steps:
+    - name: Check out code
+      uses: actions/checkout@v2
+
+    - name: Install dependencies
+      run: pip install tox
+
+    - name: Lint code
+      run: tox -vve lint
+
+  unit-test:
+    name: Unit
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v2
+
+      - name: Install dependencies
+        run: pip install tox
+
+      - name: Run unit tests
+        run: tox -e unit
+
+      - name: Upload coverage artifacts
+        uses: actions/upload-artifact@v2
+        with:
+          name: coverage
+          path: training-operator-cov_html
+        if: failure()
+
+  integration-test:
+    name: Integration
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v2
+
+      - name: Setup operator environment
+        uses: charmed-kubernetes/actions-operator@main
+        with:
+            provider: microk8s
+            channel: 1.21/stable
+            charmcraft-channel: latest/candidate
+
+      - name: Run integration tests
+        run: tox -e integration -- --model testing
+
+      - name: Capture k8s resources on failure
+        run: |
+          set -eux
+          kubectl get all -A
+          kubectl get pods -n testing --show-labels
+          kubectl get crds
+        if: failure()
+
+      - name: Get juju status
+        run: juju status
+        if: failure()
+
+      - name: Get workload logs
+        run: kubectl logs --tail 100 -ntesting -lapp.kubernetes.io/name=training-operator -ccharm
+        if: failure()
+
+      - name: Get operator logs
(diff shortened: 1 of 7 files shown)


.github/workflows/integration.yaml
```diff
@@ -0,0 +1,81 @@
...
+      # https://github.com/charmed-kubernetes/actions-operator/issues/27
+      - name: Install charmcraft latest/candidate
+        run: sudo snap refresh charmcraft --channel latest/candidate
+
+      - name: Run integration tests
+        run: tox -e integration -- --model testing
+
+      # On failure, capture debugging resources
+      - name: Get all
+        run: |
```
knkski, line 65: Multiline scripts like this should probably have `set -eux` at the beginning, it makes it easier to debug when something fails.

.github/workflows/integration.yaml
```diff
@@ -0,0 +1,81 @@
...
+        uses: actions/checkout@v2
+
+      - name: Setup operator environment
+        uses: charmed-kubernetes/actions-operator@main
+        with:
+            provider: microk8s
+            channel: 1.21/stable
+
+      # https://github.com/charmed-kubernetes/actions-operator/issues/27
+      - name: Install charmcraft latest/candidate
```
knkski, line 57: Stylistic choice, but for steps that simply run a single command, I prefer not adding a `name` field, as the default name is pretty good, and it's one more place for code and comments to get out of sync. As an example, if we skipped the `name` field on this, it would show up as `Run sudo snap refresh charmcraft --channel latest/candidate`.

.github/workflows/integration.yaml
```diff
@@ -0,0 +1,81 @@
...
+        run: sudo snap refresh charmcraft --channel latest/candidate
+
+      - name: Run integration tests
+        run: tox -e integration -- --model testing
+
+      # On failure, capture debugging resources
+      - name: Get all
```
knkski, line 64: Another stylistic choice, but I think we could just name this step `Capture debugging resources on failure` and then skip the comment.

.github/workflows/integration.yaml
```diff
@@ -0,0 +1,81 @@
...
+      - name: Check out code
+        uses: actions/checkout@v2
+
+      - name: Install dependencies
+        run: pip install tox
+
+      - name: Run unit tests
+        run: tox -e unit
+
+      # Upload coverage report on failure
```
knkski, line 35: IMO the comment is superfluous

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,46 @@
...
+    - name: Install dependencies
+      run: |
+        set -eux
+        sudo snap install charmcraft --classic --channel=latest/candidate
+        sudo snap install yq
+
+    - name: Charm
+      env:
+        CHARMCRAFT_CREDENTIALS: ${{ secrets.CHARMCRAFT_CREDENTIALS }}
+        CHARMCRAFT_DEVELOPER: "y"
```
knkski, line 29: We can actually remove this line. I forgot to remove it from `dex-auth`, but it's only necessary if you've installed `charmcraft` via pip.

tox.ini
```diff
@@ -1,22 +1,29 @@
+[flake8]
+max-line-length = 100
+ignore = W503
```
knkski, line 3: Do we actually have any W503 violations here? It doesn't look like it at a glance.

.github/workflows/integration.yaml
DnPlas: +1, though I wanted to left it very evident so we could remove it later. Now that we have [PR#13](https://github.com/charmed-kubernetes/actions-operator/pull/28) I will remove this block.

.github/workflows/integration.yaml
DnPlas: +1

.github/workflows/integration.yaml
DnPlas: +1 on both comments. Will change it.

.github/workflows/publish.yaml
DnPlas: +1, makes sense.

tox.ini
DnPlas: Not at all, will remove it.

DomFleischmann (approved):

PR #15, opened 2021-11-25
Adds aggregated ClusterRoles for multi-user usecases

### LINKED ISSUE #14: Include aggregated ClusterRoles from upstream manifests
Here are some ClusterRoles that are necessary for `training-operator` to work properly with a multi-user Kubeflow deployment:

https://github.com/kubeflow/manifests/blob/54aa7ae/apps/training-operator/upstream/overlays/kubeflow/kubeflow-training-roles.yaml

Right now, `training-operator` doesn't create them. Since it's a sidecar charm with full Kubernetes API access vs a pod spec charm, it can create them itself.

DnPlas:
Include aggregated ClusterRoles from upstream manifests.

Fixes #14

### DIFF
--- a/src/auth_manifests.yaml
+++ b/src/auth_manifests.yaml
@@ -62,3 +62,79 @@ subjects:
 - kind: ServiceAccount
   name: {{ app_name }}
   namespace: {{ namespace }}
+
+---
+apiVersion: rbac.authorization.k8s.io/v1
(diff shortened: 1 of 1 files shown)


DomFleischmann (approved):
LGTM, @ca-scribner can you also take a look?

PR #17, opened 2021-12-01
workflows/publish.yaml: use canonical/charmhub-upload-action@0.1.0 for publish job

DnPlas:
Using the Github [action](https://github.com/canonical/charmhub-upload-action) instead of manually publishing the charm.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -5,41 +5,20 @@ on:
     branches:
       - main
       - track/**
-  schedule:
-    - cron: '0 8 * * THU'
-
+      - dnplas/** #DEBUG only, will remove 
+  pull_request:
+    types:
+      - opened
+      - edited
+    branches:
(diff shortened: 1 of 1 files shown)


ca-scribner:
I don't think we want to schedule the publish workflow.  I can't remember which PR we discussed this on, but the tl/dr was that if we schedule a publish, any inherited breakage will make it into the published charms without us knowing (eg: if a dependency change its API, we will publish a new charm with the broken dependency and wont know until someone tries to use it).

DnPlas:
> I don't think we want to schedule the publish workflow. I can't remember which PR we discussed this on, but the tl/dr was that if we schedule a publish, any inherited breakage will make it into the published charms without us knowing (eg: if a dependency change its API, we will publish a new charm with the broken dependency and wont know until someone tries to use it).

Sounds reasonable. I'll remove the schedule.

PR #18, opened 2021-12-03
workflows/publish.yaml: use canonical/charmhub-upload-action@0.1.0 for publish job

DnPlas:
Using the Github [action](https://github.com/canonical/charmhub-upload-action) instead of manually publishing the charm.


DnPlas:
This change will be tracked in #19

PR #22, opened 2021-12-14
Modifies resource creation and adds defer condition 

DnPlas:
Modifying the resource creation to allow catching 403 codes,
which at the same time allows a more intelligent recovery after
a BlockedStatus by deferring the OnInstall event handler.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -112,7 +112,7 @@ class TrainingOperatorCharm(CharmBase):
             else:
                 raise
 
-    def _on_install(self, _):
+    def _on_install(self, event):
         """Event handler for InstallEvent."""
 
         # Update Pebble configuration layer if it has changed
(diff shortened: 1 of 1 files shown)


src/charm.py
```diff
@@ -121,14 +146,17 @@ def _on_install(self, _):
...
+            self._create_auth()
             self.unit.status = MaintenanceStatus("Creating CRDs")
             self._create_crds()
         except ApiError as e:
             logging.error(traceback.format_exc())
             self.unit.status = BlockedStatus(
                 f"Creating/patching resources failed with code {str(e.status.code)}."
             )
+            if e.status.code == 403:
+                event.defer()
```
ca-scribner, line 158: I don't have experience with `.defer()` so I don't know when is a good/bad time to use it.  The awkward part is that it just reruns this event again immediately before whatever the next event is, when that event happens.  

I think that doing a `juju trust` probably triggers at least a config-changed hook, so you might not need this at all.

ca-scribner (approved):
lgtm, except for how I don't know if this is an appropriate use of `defer`.  I say we merge this now just to get things working better, but check with someone like @DomFleischmann or others on whether `defer()` is good here

src/charm.py
DnPlas: > The awkward part is that it just reruns this event again immediately before whatever the next event is

In this case this behaviour is what we want because no other event handler creates the resources, waiting for config-changed will only cause more issues, like `404s` because it will try to patch something that doesn't exist. I decided to use `defer()` to avoid adding more logic to the `config_changed` event handler like: patch if exists, if not, create.

ca-scribner (approved):
lgtm with same comment as previous review

PR #27, opened 2022-03-28
Integration prometheus grafana

natalian98:
- Added integration with prometheus and grafana
- Changed raise_on_blocked=True to False due to integration tests failing - charm gets blocked during deployment but then turns active without intervention


src/charm.py
```diff
@@ -35,13 +52,24 @@ def __init__(self, *args):
...
+
         self.framework.observe(self.on.install, self._on_install)
         self.framework.observe(self.on.config_changed, self._on_config_changed)
         self.framework.observe(
             self.on.training_operator_pebble_ready,
             self._on_training_operator_pebble_ready,
         )
 
+        monitoring_events = [
+            self.on["monitoring"].relation_changed,
```
DnPlas, line 65: Shall we agree on a naming convention? I see `metrics-endpoint` and `monitoring`. Any suggestions here @simskij ?

tests/integration/test_charm.py
```diff
@@ -31,7 +36,7 @@ async def test_build_and_deploy(ops_test: OpsTest):
         charm_under_test, resources=resources, application_name=APP_NAME, trust=True
     )
     await ops_test.model.wait_for_idle(
-        apps=[APP_NAME], status="active", raise_on_blocked=True, timeout=60 * 10
+        apps=[APP_NAME], status="active", raise_on_blocked=False, timeout=60 * 10
```
DnPlas, line 39: Is there a condition where the charm blocks and then unblocks itself that we don't want to catch during testing?

tests/integration/test_charm.py
```diff
@@ -109,11 +114,42 @@ def assert_job_status_running_success():
...
+    prometheus = "prometheus-k8s"
+    grafana = "grafana-k8s"
+
+    await ops_test.model.deploy(prometheus, channel="latest/beta")
+    await ops_test.model.deploy(grafana, channel="latest/beta")
+    await ops_test.model.add_relation(prometheus, grafana)
+    await ops_test.model.add_relation(APP_NAME, grafana)
+    await ops_test.model.add_relation(prometheus, APP_NAME)
+
+    await ops_test.model.wait_for_idle(status="active", timeout=60 * 10)
```
DnPlas, line 138: I probably want to see [this](https://github.com/canonical/envoy-operator/pull/12/files#diff-08291d7f4ad7b2caecacb6ae26af540ebbaede9fe42d5e8e1b39d2bdefe4c2d4R47) to continue with the tests.

tests/integration/test_charm.py
```diff
@@ -109,11 +114,42 @@ def assert_job_status_running_success():
...
+        "address"
+    ]
+    logger.info(f"Prometheus available at http://{prometheus_unit_ip}:9090")
+
+    r = requests.get(
+        f'http://{prometheus_unit_ip}:9090/api/v1/query?query=up{{juju_application="{APP_NAME}"}}'
+    )
+    response = json.loads(r.content.decode("utf-8"))
+    response_status = response["status"]
+    logger.info(f"Response status is {response_status}")
```
DnPlas, line 151: Probably assert on the response is a good idea to catch the response we want.

DnPlas (commented):
Few comments, thanks @natalian98 ! It is great to see this implemented.

tests/integration/test_charm.py
natalian98: That happens if the training-operator was previously deployed in a controller and then re-created. Unit gets blocked with message `(leader-elected) Creating/patching resources failed with code 409` because the `kubeflow-training-admin` ClusterRole wasn't deleted along with the previous deployment. Lightkube raises this status due to leftovers and that causes the `install` hook to put the unit into `BlockedStatus`. But then the `config-changed` hook runs and patches the resources which makes the unit active without manual intervention.
This exception was handled for crds, but not for auth resources to which the ClusterRole belongs - I added it and reverted to `raise_on_blocked=True` in the integration test.

src/charm.py
natalian98: Changed to `metrics-endpoint` as discussed on different PRs

tests/integration/test_charm.py
ca-scribner: This sounds like we have a separate problem that not everything deployed by training operator gets cleaned up when juju removes the application - are there things we can do to fix that root cause?  Or is it a juju bug we can open?

DomFleischmann (approved):

PR #33, opened 2022-07-13
docs: adds CONTRIBUTING.md file

DnPlas:
Adds CONTRIBUTING.md file. It is based on the template-operator [CONTRIBUTING.md file](https://github.com/canonical/template-operator/blob/main/CONTRIBUTING.md), with instructions on how to upgrade src and resources for the operator.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,80 @@
+# Contributing
+
+## Overview
+
+This document explains the processes and practices recommended for contributing enhancements to
+this operator.
+
+- Generally, before developing enhancements to this charm, you should consider [opening an issue
+  ](https://github.com/canonical/training-operator/issues) explaining your use case.
+- If you would like to chat with us about your use-cases or proposed implementation, you can reach
+  us at [Canonical Mattermost public channel](https://chat.charmhub.io/charmhub/channels/charm-dev)
+  or [Discourse](https://discourse.charmhub.io/).
+- Familiarising yourself with the [Charmed Operator Framework](https://juju.is/docs/sdk) library
+  will help you a lot when working on new features or bug fixes.
+- All enhancements require review before being merged. Code review typically examines
+  - code quality
+  - test coverage
+  - user experience for Juju administrators of this charm.
+- Please help us out in ensuring easy to review branches by rebasing your pull request branch onto
+  the `main` branch. This also avoids merge commits and creates a linear Git commit history.
+
+## Developing
+
+You can use the environments created by `tox` for development:
+
+```shell
+tox --notest -e unit
+source .tox/unit/bin/activate
+```
+
+### Testing
+
+```shell
+tox -e lint          # code style
+tox -e unit          # unit tests
+tox -e integration   # integration tests
+tox                  # runs 'lint' and 'unit' environments
+```
+
+## Build charm
+
+Build the charm in this git repository using:
+
+```shell
+charmcraft pack
+```
+
+### Deploy
+
+```bash
+# Create a model
+juju add-model dev
+# Enable DEBUG logging
+juju model-config logging-config="<root>=INFO;unit=DEBUG"
+# Deploy the charm
+juju deploy ./training-operator_ubuntu-20.04-amd64.charm \
+    --resource training-operator-image=$(yq '.resources."training-operator-image"."upstream-source"' metadata.yaml)
+```
+
+## Updating the charm for new versions of the workload
+
+To upgrade the source and resources Charmed Training Operator, you must:
+
+1) Bump the `training-operator-image` in `metadata.yaml`
+2) Update the charm source for any changes, such as:
+
+* YAML manifests in src/ and/or any Kubernetes resource in pod_spec
+* New or changed configurations passed to pebble workloads or through pod.set_spec
+
+3) Ensure integration and unit tests are passing; fix/adapt them otherwise
+
+The following table collects a series of links and information that can help in the upgrade process.
+ 
+| Image registry 	| Image tag format	| Manifest location 	| Next version 	|
+|---	|---	|---	|---	|
+| [kubeflow/training-operator](https://hub.docker.com/r/kubeflow/training-operator/) 	| `v1-<commit sha>`<br>The commit sha comes from the HEAD of the release tag 	| [kubeflow/manifests](https://github.com/kubeflow/manifests/tree/master/apps/training-operator/upstream) 	| Check for latest version released or pre-released in <br>the repo [releases](https://github.com/kubeflow/training-operator/releases), it should match the release plan usually stated in issues like <br>[Training Operator WG and Kubeflow 1.6](https://github.com/kubeflow/manifests/issues/2196) release and [KF Release 1.6 Tracking](https://github.com/kubeflow/manifests/issues/2194) 	|
+
+## Canonical Contributor Agreement
+
+Canonical welcomes contributions to the Charmed Training Operator. Please check out our [contributor agreement](https://ubuntu.com/legal/contributors) if you're interested in contributing to the solution.


CONTRIBUTING.md
```diff
@@ -0,0 +1,76 @@
...
+### Deploy
+
+```bash
+# Create a model
+juju add-model dev
+# Enable DEBUG logging
+juju model-config logging-config="<root>=INFO;unit=DEBUG"
+# Deploy the charm
+juju deploy ./training-operator_ubuntu-20.04-amd64.charm \
+    --resource training-operator-image=kubeflow/training-operator
```
ca-scribner, line 57: Should we add a note here to remind people to add a tag to this image?  Either as a comment, or maybe something like `=kubeflow/training-operator:TAG_IN_MANIFEST`?  Otherwise, who knows whether they're getting the image we expect.  You could also just write here `kubeflow/training-operator:v1-8c6eab2`, but it'll be annoying to need to fix every time we bump versions.

An alternative could be something like: 
```
juju deploy ./training-operator_ubuntu-20.04-amd64.charm \
    --resource oci-image=$(yq eval ".resources.oci-image.upstream-source" metadata.yaml)
```

That way the command is always valid, but no maintenance.  It's kinda non-intuitive to the reader, though

CONTRIBUTING.md
```diff
@@ -0,0 +1,76 @@
...
+
+To upgrade the source and resources Charmed Training Operator, you must:
+
+1) Bump the `training-operator-image` in `metadata.yaml`
+2) Update all manifests in `src/` to match those in the version you want to upgrade to
+3) Ensure integration and unit tests are passing; fix/adapt them otherwise
+
+The following table collects a series of links and information that can help in the upgrade process.
+ 
+| Image repository 	| Image version 	| Manifest location 	| Next version 	|
```
ca-scribner, line 70: I wasn't sure what the `image version` column is supposed to convey.  Is this the format of how they tag their images?

CONTRIBUTING.md
```diff
@@ -0,0 +1,76 @@
...
+# Create a model
+juju add-model dev
+# Enable DEBUG logging
+juju model-config logging-config="<root>=INFO;unit=DEBUG"
+# Deploy the charm
+juju deploy ./training-operator_ubuntu-20.04-amd64.charm \
+    --resource training-operator-image=kubeflow/training-operator
+```
+
+## Charmed Training Operator source and resources upgrade
```
ca-scribner, line 60: I generally like this, but if we're applying it to all of our charms maybe we can make it a bit more generic without losing any content.  There's a lot of mention of "Charmed Training Operator" and similar things that I think are unnecessary - the reader will know that's the operator since they're in this repo.  I'm thinking something like:

--------------------------

## Updating the charm for new versions of the workload
( not sure I like the above wording...  Previous wording felt very specific, but I think this is the procedure we got through whenever we bump the version of upstream that we're using.  It generally happens to bump resources and change source, but neither **must** occur.  Suggestions appreciated )

To update the workload charmed by this operator, you must:

1) Bump all `oci-image` resources in `metadata.yaml` to use the new desired version
2) Update the charm for any changes, such as:
  * YAML manifests in `src/` or `KubernetesResources` in pod_spec
  * new or changed configurations passed to pebble workloads or through `pod.set_spec`
3) Ensure integration and unit tests are passing; fix/adapt them otherwise

...

CONTRIBUTING.md
ca-scribner: Actually I'm not sure what storing `image repository` here adds.  Is that something that someone trying to update the charm will want?  I'd expect they'd use whatever the training operator release page said

ca-scribner: Do we want to cite here the manifests from the workload's repo, or from `kubeflow/manifests`?  I'm torn.  We usually follow `kubeflow/manifests`, although it feels funny to put that here instead of the workload's repo.  But for something like kfp, going to their repo and looking for releases is probably misleading for our purposes

DnPlas: > I wasn't sure what the `image version` column is supposed to convey. Is this the format of how they tag their images?

It is a reference to make it easier to identify which image belongs to which release.

DnPlas: > Actually I'm not sure what storing `image repository` here adds. Is that something that someone trying to update the charm will want? I'd expect they'd use whatever the training operator release page said

How do you know where to look? I always struggle when trying to find the registry where each image is hosted. I don't see that info in the [release page](https://github.com/kubeflow/training-operator/releases/tag/v1.5.0-rc.0).

CONTRIBUTING.md
DnPlas: * The tile: yeh, I wanted to be very specific because I didn't want to confuse the concept of upgrading the workload with upgrading the charm itself. I like your proposed title.
* Instructions: agree on 1, for 2 I would change it slightly:
    2. Update the charm source for any changes, such as:

    * YAML manifests in `src/` and/or any Kubernetes resource in pod_spec
    * new or changed configurations passed to pebble workloads or through pod.set_spec

ca-scribner: lgtm!

CONTRIBUTING.md
ca-scribner: I should have put these three threads as comments on separate lines of the table :p  I'm splitting them up...

Re image version column:
I'm not sure what someone coming to this repo gains by having this info.  I'm picturing them going to the upstream manifest, looking for the images, and seeing there's new tags on the end.  I don't know what they'd do with the format of those tags beforehand?  

I have a feeling I might be misunderstanding something.

CONTRIBUTING.md
```diff
@@ -0,0 +1,76 @@
...
+
+1) Bump the `training-operator-image` in `metadata.yaml`
+2) Update all manifests in `src/` to match those in the version you want to upgrade to
+3) Ensure integration and unit tests are passing; fix/adapt them otherwise
+
+The following table collects a series of links and information that can help in the upgrade process.
+ 
+| Image repository 	| Image version 	| Manifest location 	| Next version 	|
+|---	|---	|---	|---	|
+| [kubeflow/training-operator](https://hub.docker.com/r/kubeflow/training-operator/) 	| `v1-<commit sha>`<br>The commit sha comes from the HEAD of the release tag 	| [kubeflow/training-operator/manifests](https://github.com/kubeflow/training-operator/tree/v1.5.0-rc.0/manifests) 	| Check for latest version released or pre-released in <br>the repo [releases](https://github.com/kubeflow/training-operator/releases), it should match the release plan usually stated in issues like <br>[Training Operator WG and Kubeflow 1.6](https://github.com/kubeflow/manifests/issues/2196) release and [KF Release 1.6 Tracking](https://github.com/kubeflow/manifests/issues/2194) 	|
```
ca-scribner, line 72: ### Re `image repository` column:

I think now I might understand some of our difference both for this and `Image version`... When you're looking to update Training Operator, where do you look for info?  I look at [kubeflow/manifests](https://github.com/kubeflow/manifests) (if we're updating to sync with a new kubeflow), or [training operators/manifest](https://github.com/kubeflow/training-operator/tree/master/manifests) (if we're updating an application on its own, which usually we don't do, but maybe it'll happen sometimes).  And I get the feeling you're going to the container registry to get the image tag, is that correct?  Maybe that's why we're feeling differently about the info

CONTRIBUTING.md
ca-scribner: continuing the conversation about `image repository` [here](https://github.com/canonical/training-operator/pull/33/files#r922450851)

CONTRIBUTING.md
ca-scribner, line 72: re conversation about `kubeflow/manifests` vs workload repo:
> My reasoning is: kubeflow/manifests is almost always behind the workload repo, they usually tag releases (or release candidates) about two weeks after the workload release (or pre-release).  ....

Yeah I see what you mean.  I've been thinking of this as "our KF 1.x release should be in line with KF's 1.x release", and then to me `kubeflow/manifests` makes the most sense as the source of truth.  At minimum I agree sometimes we might want to deviate from that, but I'm not sure we want to deviate by default.  I worry for applications that release quickly.  There have been times in the past where some components (I think it was kfp?) did not put their newest release into the current Kubeflow (can't remember why.  might have just been a timing thing?), so I worry that if our contributing nudges people to the component's repo first it'll make an accident easier.  Configuration could also be different between the different repos, although I don't know if that actually happens much

What do think about having links here to both:
* `kubeflow/manifests/whateverDirThisAppIsIn` (direct to the dir so people don't have to search that big dir tree)
* `thisrepo/manifests`

And instructions that make it clear that we typically sync our releases with those in kubeflow/manifests, but sometimes the repo/manifests are also helpful?

CONTRIBUTING.md
ca-scribner: continuing the conversation about `kubeflow/manifests` vs workload repo [here](https://github.com/canonical/training-operator/pull/33/files#r922458659)

DnPlas: > I'm not sure what someone coming to this repo gains by having this info. I'm picturing them going to the upstream manifest, looking for the images, and seeing there's new tags on the end. I don't know what they'd do with the format of those tags beforehand?

Where would they look for? At least for me, the experience has always been time consuming for not knowing this because I do not know if there's a file with this info. If you have it, we can link it here. 
Also, it is convenient to know the format of the tag, it makes tags predictable for automation, makes it easy to catch errors if for some reason the image tag in a PR/change doesn't match what we expect, it is easier to go into the image registry and easily find images and relate them to releases, makes it easier for someone debugging differences between images.

CONTRIBUTING.md
DnPlas: > I worry for applications that release quickly.

What worries you the most about this? They always tag their releases (and RCs).

> There have been times in the past where some components (I think it was kfp?) did not put their newest release into the current Kubeflow (can't remember why. might have just been a timing thing?), so I worry that if our contributing nudges people to the component's repo first it'll make an accident easier. 

What do you mean by "put their newest release into the current Kubeflow"?

> And instructions that make it clear that we typically sync our releases with those in kubeflow/manifests, but sometimes the repo/manifests are also helpful?

This works for me.

DomFleischmann: I generally look at something like `values.yaml` in manifests to see what tag the manifests are pointing. Having a link to the registry still might be interesting. But I would rename this from `image repository` to `image registry`

CONTRIBUTING.md
DomFleischmann: +1 lets update it.

CONTRIBUTING.md
DomFleischmann: Yeah lets update that command to take the image from metadata.

CONTRIBUTING.md
DomFleischmann: I think the tag format is very interesting, I didn't know training-operator formated it this way. I would rename this column from `image version` to `image tag format`.

DomFleischmann (commented):
Just giving my 2 cents on the conversations that were started here. All in all, this looks very promising.

CONTRIBUTING.md
ca-scribner: Agreed that this is a charm not the bundle repo, but in practice right now how `kubeflow/manifests` applies the application is more important to us than how `repo/manifests` does.  Although they're similar, they're often not the same.  @DomFleischmann are you against having both?  Knowing where to find this charm in `kubeflow/manifests` is pretty valuable as it is a pretty complicated dir tree

ca-scribner: #### Re worries about applications with quick releases (@DnPlas): 

yes, the images are tagged, but if we're looking at `repo/manifests` we have no context of which tag we should look for.  Generally, we're trying to sync with what is being released in `kubeflow/manifests @ v1.X`, which is information only available in `kubeflow/manifests`

#### Re "put their newest release into the current Kubeflow": 

By that I mean a repo may have released X.Z as their newest release, but Kubeflow is still using X.Y.  Looking at `repo/manifests` doesn't tell us that, it just points us toward how to configure X.Z.

ca-scribner: Yeah I look at manifests, not the image repo.  Manifests have the image version that should be used, but also context on how to use it.  The image version alone doesn't say much

And then if I'm picking between manifests, agreed `repo/manifests` is good for manifests deploying that application, but since we're mostly deploying a suite of interconnected applications I think the `kubeflow/manifest` repo is probably more relevant to us most of the time (otherwise maybe we'll deploy a training-operator that breaks some other thing in kubeflow)

CONTRIBUTING.md
ca-scribner: I think this is hashed out more in the other threads now, but I look at a manifest for everything.  The image repo, even if I know how versions are formatted, doesn't tell me which version I should use.  That info is in the manifests

Agreed if we're building automation around the image, knowing the format of the tag is critical.  But we aren't doing that right now, so I don't think that's a good justification on its own to bother now.  If we automated later, we'd do the work then

CONTRIBUTING.md
DomFleischmann: I like the idea of putting both `kubeflow/manifests` and `repo/manifests`, let's do that and resolve this conversation.

CONTRIBUTING.md
DomFleischmann: Given that we are sticking with linking both manifest repositories, I don't think having this column hurts. Let's keep it in and resolve this.

DomFleischmann (approved):

ca-scribner (approved):

PR #46, opened 2022-08-29
use chisme update_layer

agathanatasha:
Update training-operator to use the `update_layer` function in chisme, updated tests to match
CI would fail until [this pr](https://github.com/canonical/charmed-kubeflow-chisme/pull/25) is merged and published. Tested locally with `git+https://github.com/canonical/charmed-kubeflow-chisme@KF-605/update-layer-handler` in `requirements.txt`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ build/
 __pycache__
 .idea
 .tox
+.coverage
--- a/src/charm.py
+++ b/src/charm.py
@@ -8,6 +8,8 @@ from pathlib import Path
 
 from charms.grafana_k8s.v0.grafana_dashboard import GrafanaDashboardProvider
 from charms.prometheus_k8s.v0.prometheus_scrape import MetricsEndpointProvider
+from charmed_kubeflow_chisme.exceptions import ErrorWithStatus
+from charmed_kubeflow_chisme.pebble import update_layer
 from ops.main import main
 from ops.pebble import Layer
 from ops.charm import CharmBase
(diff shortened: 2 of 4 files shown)


src/charm.py
```diff
@@ -151,7 +138,16 @@ def _on_install(self, event):
...
+        try:
+            update_layer(
+                self._name, self._container, self._training_operator_layer, logger
+            )
+        except ErrorWithStatus as e:
+            self.model.unit.status = e.status
+            if isinstance(e.status, BlockedStatus):
+                self.logger.error(str(e.msg))
+            else:
+                self.logger.info(str(e.msg))
```
DnPlas, line 150: Let's better use `logger.error` if we are logging the message from an exception.

.gitignore
```diff
@@ -2,3 +2,4 @@ build/
 __pycache__
 .idea
 .tox
+.coverage
```
DnPlas, line 5: Let's be careful with `.*ignore` files, they might make the diff confusing. You'd want to add these files if you are changing something in the CI or in the structure of the repo. You can keep it here for now, let's avoid this in future PRs.

DnPlas (commented):
Thanks @agathanatasha, your implementation is fine, I don't have much to say about it, though I am thinking if these `try/catch` blocks that we are repeating can live in another helper method (I know we are trying to move away from that), as it seems redundant. What are your thoughts on that?

.gitignore
agathanatasha: noted. I think this file is created from running unit test. I wanted to avoid committing it. It is not currently committed.

agathanatasha:
> Thanks @agathanatasha, your implementation is fine, I don't have much to say about it, though I am thinking if these `try/catch` blocks that we are repeating can live in another helper method (I know we are trying to move away from that), as it seems redundant. What are your thoughts on that?

Not for now. I can't tell whether it would be reused more or less at this point in time. We can create a helper method in the future if it would be reused more.

PR #51, opened 2022-11-29
chore: bump all charm libs

ca-scribner:
bump all charm libs to most recent versions


DnPlas:
Thanks @ca-scribner, this change does not apply anymore, libs have been updated in another PR.

Issue #63, opened 2023-01-13
Missing on delete cleanup functionality


misohu:
After the charm is deleted crds and other manifest are not cleaned. Detailed problem description in: https://github.com/canonical/bundle-kubeflow/issues/525

NohaIhab:
fixed by #64

PR #64, opened 2023-01-16
Add remove hook + rewrite with chisme

misohu:
Fix for this issue: https://github.com/canonical/training-operator/issues/63

TLDR: on remove leaves some CRDs in cluster. Add remove hook 

Bonus: rewrite with chisme


src/charm.py
```diff
@@ -83,119 +129,78 @@ def _training_operator_layer(self) -> Layer:
...
-        # Update Pebble configuration layer if it has changed
-        self._update_layer()
-
-        # Patch/create Kubernetes resources
+        self.main(_)
+
+    def _on_install(self, _):
+        """Perform installation only actions."""
+        self._check_container_connection()
+        self._on_pebble_ready(_)
```
i-chvets, line 190: Looks good. Just a comment: Install event happens before container is started. So, chances are that container is not up at this point. With protection of `_check_container_connection()` we should be good calling Pebble Ready event.

Other solution would be let Pebble Ready event handler to handle items when container is ready and in Install event handler install K8S resources to speed up deployment.
Seldon install handler re-design does this in latest version. In previous version it was following same pattern as here.

misohu: Sooo leaving it for now ... we can change it later when facing depoyment slowdowns :)

i-chvets (approved):
Approved.

PR #83, opened 2023-03-15
fix: remove BlockedStatus statements in charm code

DnPlas:
This PR includes:
   
* fix: remove BlockedStatus statements in charm code
    A BlockedStatus mean a human has to manually intervene to ublock
    the unit and let it proceed. There were some conditions where no
    external intervention could ublock the unit. Adding appropriate
    logging and exceptions to catch errors and set status correctly.
    
    Partially fixes canonical/bundle-kubeflow#549

* fix: use _check_can_connect() in pebble ready event handler

Please merge after ~~#89~~ #93 , the CI (integration and publish jobs) is failing due to SDI dependencies incompatibilities.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -180,10 +180,7 @@ class TrainingOperatorCharm(CharmBase):
 
     def _on_pebble_ready(self, _):
         """Configure started container."""
-        if not self.container.can_connect():
-            raise ErrorWithStatus(
-                f"Container {self._container_name} failed to start", BlockedStatus
-            )
+        self._check_container_connection()
         self.main(_)
 
     def _on_install(self, _):
--- a/src/charm.py
+++ b/src/charm.py
@@ -147,8 +147,7 @@ class TrainingOperatorCharm(CharmBase):
             self.k8s_resource_handler.apply()
             self.crd_resource_handler.apply()
         except ApiError as e:
-            self.logger.exception(f"Failed to create K8S resources, with error: {e}")
-            raise e
+            raise ApiError("Failed to create K8S resources") from e
         self.model.unit.status = MaintenanceStatus("K8S resources created")
 
     def _update_layer(self) -> None:
@@ -162,8 +161,7 @@ class TrainingOperatorCharm(CharmBase):
                 self.logger.info("Pebble plan updated with new configuration, replaning")
                 self.container.replan()
             except ChangeError as e:
-                self.logger.exception(f"Failed to replan, with error: {e}")
-                raise e
+                raise ChangeError("Failed to replan") from e
 
     def main(self, _) -> None:
         """Perform all required actions the Charm."""
--- a/src/charm.py
+++ b/src/charm.py
@@ -131,10 +131,6 @@ class TrainingOperatorCharm(CharmBase):
 
     def _check_container_connection(self):
         """Check if connection can be made with container."""
-        # FIXME: if the container is unreachable, setting the unit
-        # to MaintenanceStatus won't have any effect in the execution
-        # which may cause issues if there is an attempt of starting
-        # the workload service or updating the pebble layer (which replans)
         if not self.container.can_connect():
             raise ErrorWithStatus("Pod startup is not complete", MaintenanceStatus)
 
@@ -182,7 +178,6 @@ class TrainingOperatorCharm(CharmBase):
 
     def _on_pebble_ready(self, _):
         """Configure started container."""
-        self._check_container_connection()
         self.main(_)
 
     def _on_install(self, _):
--- a/src/charm.py
+++ b/src/charm.py
@@ -5,7 +5,7 @@
 
 import logging
 
-from charmed_kubeflow_chisme.exceptions import ErrorWithStatus
+from charmed_kubeflow_chisme.exceptions import ErrorWithStatus, ErrorStatusWithMessage
 from charmed_kubeflow_chisme.kubernetes import KubernetesResourceHandler
 from charmed_kubeflow_chisme.lightkube.batch import delete_many
 from charms.grafana_k8s.v0.grafana_dashboard import GrafanaDashboardProvider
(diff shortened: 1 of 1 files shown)


i-chvets (approved):
Approved.

src/charm.py
```diff
@@ -146,8 +146,9 @@ def _deploy_k8s_resources(self) -> None:
             self.unit.status = MaintenanceStatus("Creating K8S resources")
             self.k8s_resource_handler.apply()
             self.crd_resource_handler.apply()
-        except ApiError:
-            raise ErrorWithStatus("K8S resources creation failed", BlockedStatus)
+        except ApiError as e:
+            self.logger.exception(f"Failed to create K8S resources, with error: {e}")
+            raise e
```
beliaev-maksim, line 151:
```suggestion
        except ApiError:
            self.logger.error(f"Failed to create K8S resources")
            raise
```

src/charm.py
```diff
@@ -160,8 +161,9 @@ def _update_layer(self) -> None:
             try:
                 self.logger.info("Pebble plan updated with new configuration, replaning")
                 self.container.replan()
-            except ChangeError:
-                raise ErrorWithStatus("Failed to replan", BlockedStatus)
+            except ChangeError as e:
+                self.logger.exception(f"Failed to replan, with error: {e}")
+                raise e
```
beliaev-maksim, line 166:
```suggestion
            except ChangeError:
                self.logger.error(f"Failed to replan")
                raise
```

src/charm.py
i-chvets: We should include {e} in the log message. It will help in debugging.

src/charm.py
i-chvets: We should include {e} in the log message. It will help in debugging.

beliaev-maksim: no
it will be a part of the exception

beliaev-maksim:
```
try:
    1/0
except ZeroDivisionError:
    print("divide 0")
    raise
divide 0
Traceback (most recent call last):
  File "<input>", line 2, in <module>
ZeroDivisionError: division by zero

```

src/charm.py
beliaev-maksim: see https://github.com/canonical/training-operator/pull/83#discussion_r1137107037

src/charm.py
i-chvets: The exception that is displayed in `juju status` message is too short (is there a way to make it longer?) and cuts off important details. If we add {e} to log message we will see it in `juju debug-log`. Otherwise `juju debug-log` will only show ""Failed to replan" which is kinda useless.

beliaev-maksim: there is a `raise` at the end, you will get all the traceback in the log

i-chvets: Got it.

beliaev-maksim: new update on the code snippet, idea is
```python
try: 
    1/0
except ZeroDivisionError as e:
    raise ZeroDivisionError("Do not do it") from e
```

```
Traceback (most recent call last):
  File "<input>", line 2, in <module>
ZeroDivisionError: division by zero
The above exception was the direct cause of the following exception:
Traceback (most recent call last):
  File "<input>", line 4, in <module>
ZeroDivisionError: Do not do it
```

beliaev-maksim: then juju error on the juju status will be crystal clear while we keep all the traceback from the original issue in the log

i-chvets: Sounds perfect.

DnPlas: @beliaev-maksim `raise ChangeError("Failed to replan") from e` won't work, it has certain arguments that would be very hard for the charm code to determine. I think this applies to any exception that we raise. For instance, `ChangeError` needs `err: str, change: 'Change'`. I think it would be much easier to just log the error message and raise the exception.

beliaev-maksim (dismissed):
looks good to me. Just be careful with such reqs.txt bump

src/charm.py
```diff
@@ -131,6 +131,10 @@ def _training_operator_layer(self) -> Layer:
 
     def _check_container_connection(self):
         """Check if connection can be made with container."""
+        # FIXME: if the container is unreachable, setting the unit
+        # to MaintenanceStatus won't have any effect in the execution
+        # which may cause issues if there is an attempt of starting
+        # the workload service or updating the pebble layer (which replans)
```
beliaev-maksim, line 137: but will not we check container connect each time when we call for container actions ?

we should check container can connect before each pebble layer update

DnPlas: yes, what I mean in my comment is that there is no effect. We call the check method, if it does not connect, it sets the unit to maintenance status, and then proceeds with the execution. In the juju sdk it is recommended we defer this event and return, which we did in the past, to ensure that anything that depends on the container being reachable was able to run.

I did not add the defer here because deferring may require us to re-design the charm code.

DnPlas: > basically, you need to defer only when you need to preserve order.

I disagree, deferring an event will prevent executing the next line after calling `defer()`, which can save us from attempting to do something in the workload container that will certainly fail.

I did not say defer is the only solution, though, my `FIXME` note just says that putting the unit in Waiting or MaintenanceStatus if the workload container cannot be reached won't have any effect and it will not stop the execution, which may cause errors.

beliaev-maksim: if you only want to exit, you can use sys.exit(0), that is fine

but defer on top of it preserves the order of the events. Considering following queue
eventA
eventB
eventC

then deferring in A, will try to always execute A before B and C. In case when you have know that the charm is not functionable until A is executed

beliaev-maksim: oh, but then we need either do:
```python
def foo():
  if container.can_connect():
    container.something_that_depends_on_container()
  else:
    model.unit.status = MaintenanceStatus
```

or if your code is deeply nested, eg, you check for connection in another function:
```python
def can_connect():
  if not container.can_connect():
    model.unit.status = MaintenanceStatus
    sys.exit(0)
  return True
```

beliaev-maksim: and in our case, we are raising, so, execution is broken, no?

beliaev-maksim:
@DnPlas should it land into main as well?

PR #89, opened 2023-03-20
build: remove all SDI dependencies

DnPlas:
SDI is not used in this charm, there is no need for SDI dependencies.

Fixes #83 CI.

PR #90, opened 2023-03-20
[backport of 01d43b6] build: remove all SDI dependencies (#89)

DnPlas:
SDI is not used in the charm, no installation needed.


ca-scribner (approved):
Approving this before tests pass.  This previously passed then was rebased on a change that does not affect tests.  No need to pass them again

PR #92, opened 2023-03-20
fix: added upgrade option

i-chvets:
Summary of changes:
- Added -U option to tox.ini


ca-scribner (approved):
lgtm.  did not wait for tests as this will not affect them

Issue #104, opened 2023-06-09
upgrade from 1.5 to 1.6 intermittently fails due to 409 conflict during k8s resource creation


ca-scribner:
Reproduce by:
```
juju deploy training-operator --channel 1.5/stable --trust
# wait to settle
juju deploy training-operator --channel 1.6/stable --trust
```

training-operator will appear in `juju status` as constantly working and in `MaintenanceStatus`, and logs will show it repeatedly trying to resolve the `pebble-ready` event but ending with a 409 conflict error.  

oddly, if we then
```
juju remove-application training-operator
juju deploy training-operator --channel 1.5/stable --trust
# wait to settle
juju deploy training-operator --channel 1.6/stable --trust
```

training-operator 1.6 deploys successfully.  Unclear if this is something to do with the event order, or some other inconsistency

ca-scribner:
This happens if our field_manager does not match that which kubernetes says last applied to the resources.  We previously tried to `force` on only the upgrade event, thinking that doing this would convert the resources to our current lightkube_manager, but for unknown reasons this does not appear to be enough.  

canonical/charmed-kubeflow-chisme#65 proposes that we always `force` during lightkube applies.  fwiw, ops-lib-manifest also [forces on all applies](https://github.com/canonical/ops-lib-manifest/blob/7b87046206b5aa37884ed90b7c36c498dc059978/ops/manifests/manifest.py#L287)

DnPlas:
This issue is likely caused by the training-operator 1.5 using [`patch()`](https://github.com/canonical/training-operator/blob/track/1.5/src/charm.py#L109) without specifying a fieldManager, and the type of patch strategy being set to `patch_type=PatchType.MERGE`. This may play a role in the issue because training-operator 1.6, uses `apply()` with a defined fieldManager. 

##### To consider
* This issue puts the unit in:
`training-operator/0* maintenance  executing  10.1.15.20 Creating K8S resources`
* Happens on pebbleReady.
* This is a candidate for `force=True` to resolve the conflict, but the pebbleReady event may happen before the upgrade (?) causing conflicts

Adding some logs and my steps to reproduce:

Steps to reproduce:
1. Deploy kubeflow 1.6/stable
2. Wait for training-operator to be active and idle
3. `juju refresh training-operator 1.6/stable`

Logs:
```
[...]

PR #105, opened 2023-06-09
fix: Use force on all client.apply() calls

ca-scribner:
This adds `force=True` to app lightkube.Client.apply() calls due to issues with conflicting resources, like described [here](https://github.com/canonical/training-operator/issues/104).  This does work around the issue in #104, but likely we should keep #104 open until we have a better standard solution. 

### Testing instructions
(same as #104)
```
charmcraft pack
juju deploy training-operator --channel 1.5/stable --trust
# wait to settle
juju refresh training-operator --path ./training-operator_ubuntu-20.04-amd64.charm  --resource training-operator-image=kubeflow/training-operator:v1-66aa635
```

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -194,7 +194,11 @@ class TrainingOperatorCharm(CharmBase):
             except ChangeError as e:
                 raise GenericCharmRuntimeError("Failed to replan") from e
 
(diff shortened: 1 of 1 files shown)


i-chvets (approved):
Approved.

PR #107, opened 2023-06-12
refresh charm dependencies

beliaev-maksim:
* remove upper boundary for lightkube
* update tox file
* fix integration tests


phoevos (approved):
LGTM!
Just a small comment WRT the PR description:
* Let's add a note on why we're unpinning lightkube in the first place
* We're also downgrading Juju here.
  Ideally, this should be a separate PR, but we could at least include it in the PR description and title, so that it's not buried among a bunch of other requirements changes.

PR #130, opened 2023-09-07
fix: add training-operator aggregation ClusterRoles

### LINKED ISSUE #108: Default role kubeflow-edit not permitted to create/update pytorchjob
Hello

We have this case (sf # 00361974) raised by Xperi where Profile owner is trying to create pipelines with pytorchjobs and getting below permission issue with default sa/default-editor (bound to default role/kubeflow-edit):

```
{
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {},
  "status": "Failure",
  "message": "pytorchjobs.kubeflow.org is forbidden: User \"system:serviceaccount:maria-schumacher:default-editor\" cannot create resource \"pytorchjobs\" in API group \"kubeflow.org\" in the namespace \"maria-schumacher\"",
  "reason": "Forbidden",
  "details": {
    "group": "kubeflow.org",
    "kind": "pytorchjobs"
  },
  "code": 403
}

```
I checked the default permission charts on kf docs:

https://archive-docs.d2iq.com/dkp/kaptain/1.2.0/user-management/#permissions-charts

and none of the default kubeflow roles have create/update permissions for pytorchjobs.

We can add the permission to the default kubeflow-edit role as a quick fix (using [1]) but I wanted to understand & recommend them the best practices around the RBAC in kubeflow.

Should we ask them to edit existing kubeflow-edit role, considering the user concerned in this scenario is a profile owner ?

Should we ask them to create a custom role using a copy of existing kubeflow-edit role and bind this custom role to the user ?

Refrences:
[1] https://archive-docs.d2iq.com/dkp/kaptain/1.2.0/user-management/#adding-permissions-for-a-kubeflow-user

Thanks & regards
Kamal Bhaskar

DnPlas:
Adds Aggregation ClusterRoles for training-operators so the `default-editor` can get access to manage the various *Jobs the training-operator offers. This fixes #108 in `main`, thus `latest/edge`, a follow up PR will be sent to fix the issue in the `1.7/stable` channel.

Part of #108

TODO:
- [x] Add an integration test case for covering the scenario described in #108

### DIFF
--- a/tests/integration/test_charm_with_profile.py
+++ b/tests/integration/test_charm_with_profile.py
@@ -70,7 +70,7 @@ async def test_authorization_for_creating_resources(
         check=True,
         fail_msg="Failed to execute kubectl auth",
     )
-    assert stdout == "yes"
+    assert stdout.strip() == "yes"
 
 
 def apply_manifests(lightkube_client: lightkube.Client, yaml_file_path: str):
--- a/tests/integration/test_charm_with_profile.py
+++ b/tests/integration/test_charm_with_profile.py
@@ -115,7 +115,11 @@ def apply_profile(lightkube_client):
 
     # Allow time for the Profile to be created
     try:
-        for attempt in Retrying(stop=(stop_after_attempt(10) | stop_after_delay(30)), wait=wait_exponential(multiplier=1, min=5, max=10), reraise=True):
+        for attempt in Retrying(
+            stop=(stop_after_attempt(10) | stop_after_delay(30)),
+            wait=wait_exponential(multiplier=1, min=5, max=10),
+            reraise=True,
+        ):
             with attempt:
                 lightkube_client.get(profile_resource, name=PROFILE_NAME)
     except RetryError:
--- a/tests/integration/test_charm_with_profile.py
+++ b/tests/integration/test_charm_with_profile.py
@@ -60,7 +60,7 @@ async def test_authorization_for_creating_resources(
     # Set up for creating an object of kind *Job
     job_yaml = yaml.safe_load(Path(example).read_text())
     training_job = job_yaml["kind"]
-    _, stdout, __ = await ops_test.run(
+    _, stdout, _ = await ops_test.run(
         "kubectl",
         "auth",
         "can-i",
--- a/tests/integration/test_charm_with_profile.py
+++ b/tests/integration/test_charm_with_profile.py
@@ -46,7 +46,7 @@ async def test_build_and_deploy(ops_test: OpsTest):
     )
 
     await ops_test.model.wait_for_idle(
-        status="active", raised_on_blocked=True, raise_on_error=True, timeout=60 * 10
+        status="active", raise_on_blocked=True, raise_on_error=True, timeout=60 * 10
     )
 
 
--- a/tests/integration/test_charm_with_profile.py
+++ b/tests/integration/test_charm_with_profile.py
@@ -10,7 +10,7 @@ import yaml
 from lightkube import codecs
 from lightkube.generic_resource import create_global_resource
 from pytest_operator.plugin import OpsTest
-from tenacity import RetryError, Retrying, stop_after_attempt, stop_after_delay
+from tenacity import RetryError, Retrying, stop_after_attempt, stop_after_delay, wait_exponential
 
 basedir = Path("./").absolute()
 PROFILE_NAMESPACE = "profile-example"
@@ -115,7 +115,7 @@ def apply_profile(lightkube_client):
 
     # Allow time for the Profile to be created
     try:
-        for attempt in Retrying(stop=(stop_after_attempt(5) | stop_after_delay(30)), reraise=True):
+        for attempt in Retrying(stop=(stop_after_attempt(10) | stop_after_delay(30)), wait=wait_exponential(multiplier=1, min=5, max=10), reraise=True):
             with attempt:
                 lightkube_client.get(profile_resource, name=PROFILE_NAME)
     except RetryError:
--- a/tests/integration/test_charm_with_profile.py
+++ b/tests/integration/test_charm_with_profile.py
@@ -35,11 +35,20 @@ async def test_build_and_deploy(ops_test: OpsTest):
 
     # Deploy kubeflow-roles and kubeflow-profiles to create a Profile
     await ops_test.model.deploy(
-        apps=["kubeflow-roles", "kubeflow-profiles"],
+        entity_url="kubeflow-roles",
+        channel="latest/edge",
         status="active",
         raised_on_blocked=True,
         timeout=60 * 10,
     )
+    await ops_test.model.deploy(
+        entity_url="kubeflow-profiles",
+        channel="latest/edge",
+        status="active",
+        raised_on_blocked=True,
+        timeout=60 * 10,
+    )
+
     await ops_test.model.wait_for_idle(
         status="active", raised_on_blocked=True, raise_on_error=True, timeout=60 * 10
     )
--- a/tests/integration/test_charm_with_profile.py
+++ b/tests/integration/test_charm_with_profile.py
@@ -1,6 +1,7 @@
 # Copyright 2023 Canonical Ltd.
 # See LICENSE file for licensing details.
 import glob
+import time
 from pathlib import Path
 
 import lightkube
@@ -115,6 +116,9 @@ def apply_profile(lightkube_client):
     # Apply Profile first
     apply_manifests(lightkube_client, PROFILE_FILE_PATH)
 
+    # Allow time for the Profile to be created
+    time.sleep(10)
+                    
     yield
 
     # Remove namespace
--- a/tests/integration/test_charm_with_profile.py
+++ b/tests/integration/test_charm_with_profile.py
@@ -118,7 +118,7 @@ def apply_profile(lightkube_client):
 
     # Allow time for the Profile to be created
     time.sleep(10)
-                    
+
     yield
 
     # Remove namespace
--- a/tests/integration/test_charm_with_profile.py
+++ b/tests/integration/test_charm_with_profile.py
@@ -1,7 +1,7 @@
 # Copyright 2023 Canonical Ltd.
 # See LICENSE file for licensing details.
 import glob
-import time
+import logging
 from pathlib import Path
 
 import lightkube
@@ -10,14 +10,24 @@ import yaml
 from lightkube import codecs
 from lightkube.generic_resource import create_global_resource
 from pytest_operator.plugin import OpsTest
+from tenacity import (
+    RetryError,
+    Retrying,
+    retry,
+    stop_after_attempt,
+    stop_after_delay,
+    wait_exponential,
+)
 
 basedir = Path("./").absolute()
 PROFILE_NAMESPACE = "profile-example"
-PROFILE_FILE_PATH = f"{basedir}/tests/integration/profile.yaml"
-PROFILE_FILE = yaml.safe_load(Path(PROFILE_FILE_PATH).read_text())
-KUBEFLOW_USER_NAME = PROFILE_FILE["spec"]["owner"]["name"]
+PROFILE_NAME = "profile-example"
+PROFILE_FILE_PATH = basedir / "tests/integration/profile.yaml"
+PROFILE_FILE = yaml.safe_load(PROFILE_FILE_PATH.read_text())
 METADATA = yaml.safe_load(Path("./metadata.yaml").read_text())
 APP_NAME = "training-operator"
+
+log = logging.getLogger(__name__)
 
 
 @pytest.mark.abort_on_fail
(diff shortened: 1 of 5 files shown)


tests/integration/test_charm_with_profile.py
```diff
@@ -0,0 +1,135 @@
...
+import pytest
+import yaml
+from lightkube import codecs
+from lightkube.generic_resource import create_global_resource
+from pytest_operator.plugin import OpsTest
+
+basedir = Path("./").absolute()
+PROFILE_NAMESPACE = "profile-example"
+PROFILE_FILE_PATH = f"{basedir}/tests/integration/profile.yaml"
+PROFILE_FILE = yaml.safe_load(Path(PROFILE_FILE_PATH).read_text())
```
phoevos, line 17: `basedir` is already a path, so you can just append strings to it and have the result still be path. Then you can update a couple of other places in the code that convert `PROFILE_FILE_PATH` to a `Path` all over again, and edit `apply_manifests` accordingly (suggestions in corresponding comments).
```suggestion
PROFILE_FILE_PATH = basedir / "tests/integration/profile.yaml"
PROFILE_FILE = yaml.safe_load(PROFILE_FILE_PATH.read_text())
```

tests/integration/test_charm_with_profile.py
```diff
@@ -0,0 +1,135 @@
...
+from pathlib import Path
+
+import lightkube
+import pytest
+import yaml
+from lightkube import codecs
+from lightkube.generic_resource import create_global_resource
+from pytest_operator.plugin import OpsTest
+
+basedir = Path("./").absolute()
```
phoevos, line 14: Does this really have to be an absolute path?

tests/integration/test_charm_with_profile.py
```diff
@@ -0,0 +1,135 @@
...
+        f"{training_job}",
+        f"--as=system:serviceaccount:{PROFILE_NAMESPACE}:default-editor",
+        f"--namespace={PROFILE_NAMESPACE}",
+        check=True,
+        fail_msg="Failed to execute kubectl auth",
+    )
+    assert stdout.strip() == "yes"
+
+
+def apply_manifests(lightkube_client: lightkube.Client, yaml_file_path: str):
```
phoevos, line 77:
```suggestion
def apply_manifests(lightkube_client: lightkube.Client, yaml_file_path: Path):
```

tests/integration/test_charm_with_profile.py
```diff
@@ -0,0 +1,135 @@
...
+    """Apply resources using manifest files and returns the applied object.
+
+    Args:
+        lightkube_client (lightkube.Client): an instance of lightkube.Client to
+            use for applying resources.
+        yaml_file_path (str): the resource yaml file.
```
phoevos, line 83:
```suggestion
        yaml_file_path (Path): the path to the resource yaml file.
```

tests/integration/test_charm_with_profile.py
```diff
@@ -0,0 +1,135 @@
...
+
+    Returns:
+        A namespaced or global lightkube resource (obj).
+    """
+    read_yaml = Path(yaml_file_path).read_text()
```
phoevos, line 88:
```suggestion
    read_yaml = yaml_file_path.read_text()
```

tests/integration/test_charm_with_profile.py
```diff
@@ -0,0 +1,135 @@
...
+    # Apply Profile first
+    apply_manifests(lightkube_client, PROFILE_FILE_PATH)
+
+    # Allow time for the Profile to be created
+    time.sleep(10)
+
+    yield
+
+    # Remove namespace
+    read_yaml = Path(PROFILE_FILE_PATH).read_text()
```
phoevos, line 125:
```suggestion
    read_yaml = PROFILE_FILE_PATH.read_text()
```

tests/integration/test_charm_with_profile.py
```diff
@@ -0,0 +1,135 @@
...
+    yaml_loaded = codecs.load_all_yaml(read_yaml)
+    for obj in yaml_loaded:
+        try:
+            lightkube_client.delete(
+                res=type(obj),
+                name=obj.metadata.name,
+                namespace=obj.metadata.namespace,
+            )
+        except lightkube.core.exceptions.ApiError as api_error:
+            raise api_error
```
phoevos, line 135: Same observation as above, can't we just let the error be raised here since we're not handling it in any special way?
```suggestion
        lightkube_client.delete(
            res=type(obj),
            name=obj.metadata.name,
            namespace=obj.metadata.namespace,
        )
```

tests/integration/test_charm_with_profile.py
```diff
@@ -0,0 +1,135 @@
...
+def lightkube_client() -> lightkube.Client:
+    """Returns a lightkube Client that can talk to the K8s API."""
+    client = lightkube.Client(field_manager="kfp-operators")
+    return client
+
+
+@pytest.fixture(scope="module")
+def apply_profile(lightkube_client):
+    """Apply a Profile simulating a user."""
+    # Create a Viewer namespaced resource
```
phoevos, line 111:
```suggestion
```

tests/integration/test_charm_with_profile.py
```diff
@@ -0,0 +1,135 @@
...
+    profile_class_resource = create_global_resource(  # noqa F841
```
phoevos, line 112: We're not using the assigned value anyways, so no need to keep it (and thus no need for the noqa)
```suggestion
    create_global_resource(
```

tests/integration/test_charm_with_profile.py
```diff
@@ -0,0 +1,135 @@
...
+        group="kubeflow.org", version="v1", kind="Profile", plural="profiles"
+    )
+
+    # Apply Profile first
+    apply_manifests(lightkube_client, PROFILE_FILE_PATH)
+
+    # Allow time for the Profile to be created
+    time.sleep(10)
```
phoevos, line 120: Why `sleep`? Do we do it like this elsewhere? Can't we have a retrying logic like in the rest of our charms? This doesn't even guarantee that the profile has been created after the 10" have passed. Wouldn't it be better to keep retrying to fetch the profile for those 10" (or a simplified version of what we do [here](https://github.com/canonical/kubeflow-profiles-operator/blob/main/tests/integration/test_charm.py))?

tests/integration/test_charm_with_profile.py
```diff
@@ -0,0 +1,135 @@
...
+import yaml
+from lightkube import codecs
+from lightkube.generic_resource import create_global_resource
+from pytest_operator.plugin import OpsTest
+
+basedir = Path("./").absolute()
+PROFILE_NAMESPACE = "profile-example"
+PROFILE_FILE_PATH = f"{basedir}/tests/integration/profile.yaml"
+PROFILE_FILE = yaml.safe_load(Path(PROFILE_FILE_PATH).read_text())
+KUBEFLOW_USER_NAME = PROFILE_FILE["spec"]["owner"]["name"]
```
phoevos, line 18: Doesn't look like we're using this anywhere.
```suggestion
```

tests/integration/test_charm_with_profile.py
```diff
@@ -0,0 +1,135 @@
...
+    read_yaml = Path(yaml_file_path).read_text()
+    yaml_loaded = codecs.load_all_yaml(read_yaml)
+    for obj in yaml_loaded:
+        try:
+            lightkube_client.apply(
+                obj=obj,
+                name=obj.metadata.name,
+            )
+        except lightkube.core.exceptions.ApiError as api_error:
+            raise api_error
```
phoevos, line 97: Since we're not doing anything with this error, can't we just let it be raised?
```suggestion
        lightkube_client.apply(
            obj=obj,
            name=obj.metadata.name,
        )
```

tests/integration/test_charm_with_profile.py
```diff
@@ -0,0 +1,135 @@
...
+        f"--as=system:serviceaccount:{PROFILE_NAMESPACE}:default-editor",
+        f"--namespace={PROFILE_NAMESPACE}",
+        check=True,
+        fail_msg="Failed to execute kubectl auth",
+    )
+    assert stdout.strip() == "yes"
+
+
+def apply_manifests(lightkube_client: lightkube.Client, yaml_file_path: str):
+    """Apply resources using manifest files and returns the applied object.
```
phoevos, line 78: nit:
```suggestion
    """Apply resources using manifest files and return the applied object.
```

tests/integration/test_charm_with_profile.py
```diff
@@ -0,0 +1,135 @@
...
+                name=obj.metadata.name,
+            )
+        except lightkube.core.exceptions.ApiError as api_error:
+            raise api_error
+    return obj
+
+
+@pytest.fixture(scope="module")
+def lightkube_client() -> lightkube.Client:
+    """Returns a lightkube Client that can talk to the K8s API."""
```
phoevos, line 103: nit:
```suggestion
    """Return a lightkube Client that can talk to the K8s API."""
```

phoevos (changes requested):
Thanks @DnPlas, left a bunch of comments with suggestions (you can commit them directly if you agree).

tests/integration/test_charm_with_profile.py
```diff
@@ -0,0 +1,135 @@
...
+@pytest.mark.parametrize("example", glob.glob("examples/*.yaml"))
+@pytest.mark.abort_on_fail
+async def test_authorization_for_creating_resources(
+    example, ops_test: OpsTest, lightkube_client, apply_profile
+):
+    """Assert a *Job can be created by a user in the user namespace."""
+    # Set up for creating an object of kind *Job
+    job_yaml = yaml.safe_load(Path(example).read_text())
+    training_job = job_yaml["kind"]
+    _, stdout, __ = await ops_test.run(
```
phoevos, line 63: Is this a mistype or is it intentional?
```suggestion
    _, stdout, _ = await ops_test.run(
```

tests/integration/test_charm_with_profile.py
```diff
@@ -0,0 +1,135 @@
...
+    charm_under_test = await ops_test.build_charm(".")
+    image_path = METADATA["resources"]["training-operator-image"]["upstream-source"]
+    resources = {"training-operator-image": image_path}
+
+    await ops_test.model.deploy(
+        charm_under_test, resources=resources, application_name=APP_NAME, trust=True
+    )
+    await ops_test.model.wait_for_idle(
+        apps=[APP_NAME], status="active", raise_on_blocked=True, timeout=60 * 10
+    )
```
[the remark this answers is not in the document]
DnPlas: We can, let me remove this wait.

tests/integration/test_charm_with_profile.py
```diff
@@ -0,0 +1,135 @@
...
+
+    # Deploy kubeflow-roles and kubeflow-profiles to create a Profile
+    await ops_test.model.deploy(
+        entity_url="kubeflow-roles",
+        channel="latest/edge",
+        trust=True,
+    )
```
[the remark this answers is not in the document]
DnPlas: Yes, roles will apply the kubeflow-edit roles that no other charm creates.

tox.ini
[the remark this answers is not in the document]
DnPlas: +1

tox.ini
```diff
@@ -73,8 +73,14 @@ deps =
...
-[testenv:integration]
-commands = pytest -v --tb native --asyncio-mode=auto {[vars]tst_path}integration --log-cli-level=INFO -s {posargs}
+[testenv:charm-integration]
+commands = pytest -v --tb native --asyncio-mode=auto {[vars]tst_path}integration/test_charm.py --log-cli-level=INFO -s {posargs}
 deps =
     -r requirements-integration.txt
 description = Run integration tests
+
+[testenv:integration-with-profiles]
+commands = pytest -v --tb native --asyncio-mode=auto {[vars]tst_path}integration/test_charm_with_profile.py --log-cli-level=INFO -s {posargs}
```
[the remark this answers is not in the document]
DnPlas: +1

tests/integration/test_charm_with_profile.py
DnPlas: It was a simplified way (the lazy way), I have updated it using tenacity and getting the Profile. Let me know what you think!

DnPlas:
@phoevos I have addressed all your comments, I believe. Ready for another round.

tests/integration/test_charm_with_profile.py
phoevos: Why though, given that we don't use them anyway.

phoevos:
Thanks @DnPlas! I still think that [this](https://github.com/canonical/training-operator/pull/130#discussion_r1319820581) is super weird but I won't block on it😋
I'll approve as soon as the CI completes!

tests/integration/test_charm_with_profile.py
DnPlas: My understanding is that is the convention, but if I'm wrong I can change it.

DnPlas: Accepted the suggestion, though, no need to have a distinction.

phoevos (approved):

PR #157, opened 2024-04-27
ci: Add Maximise GH runner space

orfeas-k:
Ref canonical/bundle-kubeflow#813 and more specifically https://github.com/canonical/bundle-kubeflow/issues/813#issuecomment-2045924903


kimwnasptd:
@orfeas-k can we add a link in the description to the issue that we had for switching to this action? Looks good otherwise

kimwnasptd (approved):

Issue #159, opened 2024-05-21
Update `training-operator` manifests


misohu:
### Context

Each charm has a set of manifest files that have to be upgraded to their target version. The process of upgrading manifest files usually means going to the component’s upstream repository, comparing the charm’s manifest against the one in the repository and adding the missing bits in the charm’s manifest.

### What needs to get done

https://docs.google.com/document/d/1a4obWw98U_Ndx-ZKRoojLf4Cym8tFb_2S7dq5dtRQqs/edit?pli=1#heading=h.jt5e3qx0jypg

### Definition of Done

1. Manifests are updated 
2. Upstream image is used

DnPlas:
The `training-operator` recently introduced `ValidatingWebhookConfigurations` (https://github.com/kubeflow/training-operator/issues/1993) for the various training resources this controller manages. Under the hood, the `training-operator` controller uses [cert-controller](https://github.com/open-policy-agent/cert-controller) for managing self-signed certificates that can be used in the webhook configuration.
The `cert-controller` component is used as part of the [business logic](https://github.com/kubeflow/training-operator/blob/master/pkg/cert/cert.go) of the `training-operator`, and in fact any potential error while managing the certificates, can cause the service to exit (see [here](https://github.com/kubeflow/training-operator/blob/master/cmd/training-operator.v1/main.go#L156C1-L160C1)).

It is important to note that the `cert-controller` will use a `Secret` as its source of truth for managing the certificates.

Based on the above information, the following were my approaches for adding support for these webhooks:

1. Using the known [`cert.py`](https://github.com/canonical/katib-operators/blob/main/charms/katib-controller/src/certs.py) script for generating self-signed certs and then using them for rendering webhooks is not an option because the actual workload controller expects the secret contents to be located in a specific volume in the training operator workload container; otherwise it will end up timing out waiting for the files. [This line](https://github.com/open-policy-agent/cert-controller/blob/master/pkg/rotator/rotator.go#L862) gets executed and the training operator service never starts correctly, more over, the webhook service never starts.
2. The `training-operator Deployment` (upstream) explicitly [mounts](https://github.com/kubeflow/training-operator/blob/v1.8-branch/manifests/base/deployment.yaml#L40C1-L43C29) the `Secret` in a `mountPath`, which is used by the `cert-controller`. Because of that, we can patch the `training-operator StatefulSet` and mount the secret in the expected `volumeMount`.
[...]

PR #160, opened 2024-05-23
refactor: remove unused ClusterRole

DnPlas:
The training-operator charm is always deployed using trust because it needs elevated permissions to apply k8s resources using lightkube, so there is not an actual need for a Role or a ClusterRole. This commit removes the unused ClusterRole, keeping just the aggregated roles (for them to be applied to the Kubeflow Profile namespace) when needed.


DnPlas:
Closing because of #159 we will be refactoring the charm and these files will be required.

PR #162, opened 2024-05-31
chore: bump training-operator v1.7 -> v1.8

### LINKED ISSUE #159: Update `training-operator` manifests

DnPlas:
This commit introduces the following changes:

* The charm now renders and applies a ValidatingWebhookConfiguration resource for 
  training-operator CRDs.
* The charm will render the Service to also serve on port 9443 for the webhook service.
* The oci-image is updated to v1.8 of the training-operator
* The training-operator Deployment now has a volume mount for mounting
  the secret that is used by the cert-controller to generate and rotate
  certificates for the ValidatingWebhookConfiguration
* The training-operator Deployment will now take an argument so the webhook service can use the training-operator workload's Service instead of the default
* Updates the examples directory with examples from kubeflow/training-operator v1.8-branch

Fixes #159


src/charm.py
```diff
@@ -61,9 +65,12 @@ def __init__(self, *args):
         self.framework.observe(self.on.remove, self._on_remove)
 
         metrics_port = ServicePort(int(METRICS_PORT), name="metrics-port")
+        webhook_port = ServicePort(
+            port=443, targetPort=9443, protocol="TCP", name="webhook-server"
```
orfeas-k, line 69:
```suggestion
            port=443, targetPort=WEBHOOK_PORT, protocol="TCP", name="webhook-server"
```
?

src/charm.py
```diff
@@ -116,6 +123,7 @@ def crd_resource_handler(self, handler: KubernetesResourceHandler):
 
     def _check_leader(self):
         """Check if this unit is a leader."""
+        self.framework.observe(self.on.update_status, self._on_event)
```
orfeas-k, line 126: Do we need this? :sweat_smile:

src/templates/validatingwebhookconfiguration.yaml.j2
```diff
@@ -0,0 +1,109 @@
+---
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingWebhookConfiguration
+metadata:
+  labels:
+    control-plane: kubeflow-training-operator
```
orfeas-k, line 6: I see upstream doesn't have that. Do we need it? Should we add a comment that this is diverging from upstream so if we need, we don't remove it in the future?

orfeas-k (commented):
Thank you @DnPlas, leaving some comments plus IIUC the following changes are missing:
* image + `secretGenerator` field https://github.com/kubeflow/manifests/pull/2699/files#diff-23b73c27a3cd384ed58c680eac804ff86be38f48688cc94742c51c530e4164e8
* changes in aggregation roles https://github.com/kubeflow/manifests/pull/2699/files#diff-8b80f589b6814910a625e07ed23ff511ef76e81cf009dad169eba0f26ed04497
* changes in training-operator role https://github.com/kubeflow/manifests/pull/2699/files#diff-a88903de8c7c015c3e145b77718f9cee059aedfc24e3b0f37d246f979a4bb15b

src/charm.py
DnPlas: Oh, haha, it must be a leftover from the rebase (it was added by mistake in another PR) :P will remove it.

src/templates/validatingwebhookconfiguration.yaml.j2
DnPlas: This label is added when we render the manifests with `kustomize build apps/training-operator/upstream/overlays/kubeflow`.

DnPlas:
https://github.com/kubeflow/training-operator/issues/2143 could be affecting (intermittently) the CI.

src/charm.py
```diff
@@ -29,7 +29,8 @@
 ]
 METRICS_PATH = "/metrics"
 METRICS_PORT = "8080"
-WEBHOOK_PORT = "9443"
+WEBHOOK_PORT = "443"
+WEBHOOK_TARGET_PORT = "9443"
```
orfeas-k, line 33: Doesn't this mess with L54? We were already using webhook_port but for 9443

.github/workflows/integrate.yaml
```diff
@@ -88,10 +88,14 @@ jobs:
         run: juju status
         if: failure()
 
-      - name: Get workload logs
-        run: kubectl logs --tail 100 -ntesting -lapp.kubernetes.io/name=training-operator -ccharm
+      - name: Get validatingwebhookconfigurations
+        run: kubectl get validatingwebhookconfigurations validator.training-operator.kubeflow.org -oyaml
```
orfeas-k, line 92: good job on imporving logs

src/templates/secret.yaml.j2
```diff
@@ -1,5 +1,5 @@
 apiVersion: v1
-data: {}
+data: {{ '{}' }}
```
orfeas-k, line 2: I trust you that this works as we want it to :D

orfeas-k (commented):
Some small comments

src/charm.py
DnPlas: Oh, let me correct it, good catch!

DnPlas:
Please review #171 before this, it should help working around the CI issue.

DnPlas:
This PR is now affected by https://github.com/canonical/training-operator/issues/174

NohaIhab (commented):
thx @DnPlas, almost there
We need to update the yamls in `/examples`, I see some images for the examples were updated in the training-opeartor [1.8 branch](https://github.com/kubeflow/training-operator/tree/v1.8-branch/examples)

NohaIhab (approved):
LGTM

Issue #170, opened 2024-06-26
`training-operator` cannot be upgraded from `1.7/stable` to recent version


DnPlas:
### Bug Description

Due to https://github.com/canonical/training-operator/pull/161, the latest `training-operator` charm will only have one container in the scheduled `Pod`, which corresponds to the container running the charm, as the workload will be deployed and scheduled separately.
Because of https://bugs.launchpad.net/juju/+bug/1991955, upgrading the charm from `1.7/stable` to the most recent version, the `Pod` will still have two containers instead of one after running `juju refresh training-operator --channel latest/edge`.

### To Reproduce

1. Deploy `juju training-operator --channel 1.7/stable --trust`
2. Refresh `juju refresh training-operator --channel latest/edge/pr-162` (this works right now) or `juju refresh training-operator --channel latest/edge` (this will only work once https://github.com/canonical/training-operator/pull/167 is merged)
3. Observe

### Environment

microk8s 1.29/stable
juju 3.5.0

### Relevant Log Output

```shell
# Getting the pods after a refresh shows the  training-operator-0 Pod has 2/2
# which has the charm and the workload container
# Then because of the refactor, we have another Pod training-operator-7f97689fcf-2zshp
# with the workload as well

$ kubectl get pods -A
testing                               training-operator-0                      2/2     Running   0          10m
testing                               training-operator-7f97689fcf-2zshp       1/1     Running   0          9m59s
```

### Additional Context

This issue is currently affecting the `test_upgrade` integration test case and the upgrade path.

##### Workarounds

So far, the only workaround that has worked is to remove the operator and re-deploy it.

mvlassis:
Closing this issue as `1.7/stable` is not in the supported versions anymore, so we will not be looking into it

misohu:
This version is no longer supported.

PR #171, opened 2024-06-26
tests: skip test_upgrade due to #170

DnPlas:
#170 is affecting the execution of this test, but since the fix is on juju, there is not much we can do at the moment other than skipping the test.

Part of #170

### DIFF
--- a/tests/integration/test_charm.py
+++ b/tests/integration/test_charm.py
@@ -181,7 +181,7 @@ async def test_remove_with_resources_present(ops_test: OpsTest):
     _last = object()
     assert next(crd_list, _last) is _last
 
-
+@pytest.mark.skip("Due to https://github.com/canonical/training-operator/issues/170")
 @pytest.mark.abort_on_fail
 async def test_upgrade(ops_test: OpsTest):
     """Test upgrade.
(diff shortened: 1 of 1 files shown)


tests/integration/test_charm.py
```diff
@@ -246,6 +247,7 @@ async def test_upgrade(ops_test: OpsTest):
             assert "paddlejobs" in rule.resources
 
 
+@pytest.mark.skip("Due to https://github.com/canonical/training-operator/issues/170")
 @pytest.mark.abort_on_fail
 async def test_remove_without_resources(ops_test: OpsTest):
     """Test remove when no resources are present.
```
NohaIhab, line 253: why should we skip this test? I think testing successful remove is still valid

NohaIhab: just the docstring needs to be updated if we decide to not skip it

DnPlas: There is another test case that removes the charm right before `test_upgrade`, if we leave this one, it will attempt to remove something that is not deployed. 

I think that in the long run these test cases have to be refactored.

NohaIhab (approved):
this LGTM, let's keep #170 in mind when testing and defining the upgrade process to CKF 1.9. I suggest to link it in the corresponding 1.9 task (let's create the upgrade task if it's not there yet). Thx!

PR #173, opened 2024-06-26
refactor: apply a workload Service instead of using juju created one

DnPlas:
To avoid inconsistent behaviours, it is preferrable to apply and use a Service owned by the charm so it can be rendered as needed by the controller.


src/templates/secret.yaml.j2
NohaIhab, line 1: name of the file was intended to be `service.yaml.j2` I think

NohaIhab (commented):
hi @DnPlas 
first, left a small comment
second, can you explain what was the issue you were seeing? what is the problem this PR is trying to solve? I'm struggling to interpret that. Would be better if it is documented in an issue so we have a reference, wdyt?

DnPlas:
> hi @DnPlas first, left a small comment second, can you explain what was the issue you were seeing? what is the problem this PR is trying to solve? I'm struggling to interpret that. Would be better if it is documented in an issue so we have a reference, wdyt?

The problem is that the Service created by juju will try to be used by two different Pods (the training operator charm and the training operator workload). This will cause inconsistencies like when we try to reach the metrics endpoint. We know that the training operator workload has it, but the charm doesn't, so sometimes trying to reach this endpoint will result in:

```
curl -v 10.152.183.20:8080/metrics
*   Trying 10.152.183.20:8080...
* connect to 10.152.183.20 port 8080 failed: Connection refused
* Failed to connect to 10.152.183.20 port 8080 after 0 ms: Connection refused
* Closing connection 0
curl: (7) Failed to connect to 10.152.183.20 port 8080 after 0 ms: Connection refused
```

To avoid having both Pods "fighting" over the same Service (which btw is not at all recommended, it goes against the logic of services), I'm proposing that the workload has its own.

src/charm.py
```diff
@@ -60,21 +59,20 @@ def __init__(self, *args):
...
         self.framework.observe(self.on.remove, self._on_remove)
 
-        metrics_port = ServicePort(int(METRICS_PORT), name="metrics-port")
-        self.service_patcher = KubernetesServicePatch(
-            self,
-            [metrics_port],
-            service_name=f"{self.model.app.name}",
-        )
-
+        # The target is the Service (applied with service.yamlj2) and the name has the following
```
NohaIhab, line 62:
```suggestion
        # The target is the Service (applied with service.yaml.j2) and the name has the following
```

NohaIhab (commented):
thx @DnPlas, tested and I was able to curl the metrics endpoint successfully. 
Small nit for a typo then I'm approving, feel free to ping me to approve.

NohaIhab (approved):
LGTM

PR #175, opened 2024-07-01
refactor: rename charm and workload services to match upstream values

### LINKED ISSUE #174: `training-operator` controller has `training-operator Service` hardcoded
### Bug Description

In a recent version of the `training-operator`, upstream introduced a `ValidatingWebhookConfiguration` which uses the `training-operator Service` as the webhook service. The name of the service is hard coded [here](https://github.com/kubeflow/training-operator/blob/master/cmd/training-operator.v1/main.go#L111).

This is problematic because the `Service` juju creates will take the name of the application, which in our case is always `training-operator`, causing the charm to own it. In recent changes (#173) we refactored the charm to apply its own `Service` to avoid using the juju created one, but it seems like that is not enough, the workload requires its service to be named `training-operator`.

### To Reproduce

1. Deploy the charm from `KF-5692-upgrade-trainig-operator-1.8` branch
2. Try applying any of the examples from `examples/`
3. See the logs

### Environment

microk8s 1.25-strict/stable
juju 3.4/stable (3.4.4)

### Relevant Log Output

```shell
$ kubectl apply -f https://raw.githubusercontent.com/canonical/training-operator/main/examples/pytorchjob.yaml
Error from server (InternalError): error when creating "https://raw.githubusercontent.com/canonical/training-operator/main/examples/pytorchjob.yaml": Internal error occurred: failed calling webhook "validator.pytorchjob.training-operator.kubeflow.org": failed to call webhook: Post "https://training-operator.to.svc:443/validate-kubeflow-org-v1-pytorchjob?timeout=10s": no service port 443 found for service "training-operator"
```

### Additional Context

_No response_

DnPlas:
Using the KubernetesServicePatch library, patch the name of the juju created/owned Service so it's different from training-operator (when the charm is deployed with that name). Also hardcode the name of the workload Service (applied with service.yaml.j2) to match upstream hardcoded value training-operator.

Fixes #174

### DIFF
--- a/lib/charms/observability_libs/v1/kubernetes_service_patch.py
+++ b/lib/charms/observability_libs/v1/kubernetes_service_patch.py
@@ -1,0 +1,400 @@
+# Copyright 2021 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""# KubernetesServicePatch Library.
+
+This library is designed to enable developers to more simply patch the Kubernetes Service created
+by Juju during the deployment of a sidecar charm. When sidecar charms are deployed, Juju creates a
+service named after the application in the namespace (named after the Juju model). This service by
+default contains a "placeholder" port, which is 65536/TCP.
+
+When modifying the default set of resources managed by Juju, one must consider the lifecycle of the
+charm. In this case, any modifications to the default service (created during deployment), will be
+overwritten during a charm upgrade.
+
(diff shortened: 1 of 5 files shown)


DnPlas:
Closing because of https://github.com/canonical/training-operator/issues/174#issuecomment-2204167475

PR #193, opened 2024-08-28
Update tools/get-images.sh to capture images properly

### LINKED ISSUE #192: `tools/get-images.sh` doesn't return any images
### Bug Description

Running the `tools/get-images.sh` returns no images for this charm. In reality, it should return the value of the "default" key in the option `training-operator-image` in `config.yaml`.

### To Reproduce

- Clone the repository with `git clone` and `cd` to it
- Run `bash tools/get-images.sh`
- The output is empty

### Environment

Branch used: `main`

### Relevant Log Output

```shell
.
```

### Additional Context

The reason is that the image for `training-operator` has been moved from `metadata.yaml` to `config.yaml` in [this PR](https://github.com/canonical/training-operator/pull/161/files).

mvlassis:
Closes #192.

This PR updates `tools/get-images.sh` to capture images properly. The bash script is updated to look at `config.yaml` instead of `metadata.yaml`, and searches all options that match the regular expression "-image$" (meaning all options that end with "image". to capture the value in the `default` key. Note that in our case only 1 value should be captured, but this way is more robust if more images are added in the future.

### DIFF
--- a/tools/get-images.sh
+++ b/tools/get-images.sh
@@ -4,6 +4,6 @@
 #
 # dynamic list
 IMAGE_LIST=()
-IMAGE_LIST+=($(find -type f -name config.yaml -exec yq '.options | with_entries(select(.key | test("-image$"))) | .[].default' {} \;))
+IMAGE_LIST+=($(find -type f -name config.yaml -exec yq '.options.*-image |  select(.) | .default' {} \;))
 printf "%s\n" "${IMAGE_LIST[@]}"
 
--- a/tools/get-images.sh
+++ b/tools/get-images.sh
@@ -4,6 +4,6 @@
 #
 # dynamic list
 IMAGE_LIST=()
-IMAGE_LIST+=($(find -type f -name metadata.yaml -exec yq '.resources | to_entries | .[] | .value | ."upstream-source"' {} \;))
+IMAGE_LIST+=($(find -type f -name config.yaml -exec yq '.options | with_entries(select(.key | test("-image$"))) | .[].default' {} \;))
 printf "%s\n" "${IMAGE_LIST[@]}"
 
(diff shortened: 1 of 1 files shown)


tools/get-images.sh
```diff
@@ -4,6 +4,6 @@
 #
 # dynamic list
 IMAGE_LIST=()
-IMAGE_LIST+=($(find -type f -name metadata.yaml -exec yq '.resources | to_entries | .[] | .value | ."upstream-source"' {} \;))
+IMAGE_LIST+=($(find -type f -name config.yaml -exec yq '.options.*-image |  select(.) | .default' {} \;))
```
rgildein, line 7: This is failing for me.
```bash
./tools/get-images.sh
jq: error: syntax error, unexpected '*', expecting FORMAT or QQSTRING_START or '[' (Unix shell quoting issues?) at <top-level>, line 1:
.options.*-image |  select(.) | .default         
jq: 1 compile error
```

mvlassis: I have created an issue for it [here](https://github.com/canonical/kfp-operators/issues/562#issue-2492407938), basically this is related to the version of `yq` used (I'm assuming you installed yq with `pip`). The solution is to use the `snap` version of `yq`.

rgildein: You are right, I had also yq installed by pip and snap. Removing version 3.1.0 installed by pip, and it's working fine :+1:

rgildein (approved):
LGTM, I tested it also on kfp-operators.

rgildein (approved):
LGTM,
tested versions:
- yq (`v4.44.2 (2566)`) installed with snap
- yq (`Version: 3.4.3`) installed with pip

DnPlas (approved):
Thanks @mvlassis, looks good.

PR #200, opened 2024-09-18
feat: Add terraform module

### LINKED ISSUE #197: Create a terraform module for training-operator charm
### Context

According to the terraform modules organization spec, this is the first step to provide a terraform solution module. We need to create a terraform module for each charm.

### What needs to get done

Create a terraform module for this charm that complies with the [spec](https://docs.google.com/document/d/1EG71A2pJ244PQRaGVzGj7Mx2B_bzE4U_OSqx4eeVI1E/edit?usp=drivesdk).

### Definition of Done

Charm has a terraform module

orfeas-k:
Create a `terraform/` directory that hosts the Terraform module for this charm. It follows the structure proposed in [this spec](https://docs.google.com/document/d/1EG71A2pJ244PQRaGVzGj7Mx2B_bzE4U_OSqx4eeVI1E/edit) and it is follows what was done in canonical/argo-operators#198.

To test the module:
- Clone the repository and switch this PR's branch.
- First run `tox -e tflint` to ensure that linting is correct
- Run `terraform apply -var "channel=latest/edge" -var "model_name=kubeflow" --auto-approve` and wait until the charm is `Active` and `Idle`.

Ref #197

### DIFF
--- a/.github/workflows/integrate.yaml
+++ b/.github/workflows/integrate.yaml
@@ -40,8 +40,6 @@ jobs:
     uses: canonical/charmed-kubeflow-workflows/.github/workflows/terraform-checks.yaml@main
     with:
       charm-path: .
-      model: kubeflow
-      channel: latest/edge
 
   integration-test:
     name: Integration
--- a/.github/workflows/integrate.yaml
+++ b/.github/workflows/integrate.yaml
@@ -35,6 +35,14 @@ jobs:
           path: training-operator-cov_html
         if: failure()
 
+  terraform-checks:
+    name: Terraform
+    uses: canonical/charmed-kubeflow-workflows/.github/workflows/terraform-checks.yaml@main
+    with:
+      charm-path: .
+      model: kubeflow
+      channel: latest/edge
+
   integration-test:
     name: Integration
     runs-on: ubuntu-20.04
--- a/terraform/README.md
+++ b/terraform/README.md
@@ -1,0 +1,63 @@
+# Terraform module for training-operator
+
+This is a Terraform module facilitating the deployment of the training-operator charm, using the [Terraform juju provider](https://github.com/juju/terraform-provider-juju/). For more information, refer to the provider [documentation](https://registry.terraform.io/providers/juju/juju/latest/docs). 
+
+## Compatibility
+This terraform module is compatible with charms of version >= 1.8 due to changes in the charm's relations.
+
+## Requirements
+This module requires a `juju` model to be available. Refer to the [usage section](#usage) below for more details.
+
+## API
+
+### Inputs
+The module offers the following configurable inputs:
+
+| Name | Type | Description | Required |
+| - | - | - | - |
+| `app_name`| string | Application name | False |
+| `channel`| string | Channel that the charm is deployed from | False |
+| `config`| map(string) | Map of the charm configuration options | False |
+| `model_name`| string | Name of the model that the charm is deployed on | True |
+| `resources`| map(number) | Map of charm resources revisions | False |
+| `revision`| number | Revision number of the charm name | False |
+
+### Outputs
+Upon applied, the module exports the following outputs:
+
+| Name | Description |
+| - | - |
+| `app_name`|  Application name |
+| `provides`| Map of `provides` endpoints |
+| `requires`|  Map of `reqruires` endpoints |
+
+## Usage
+
+This module is intended to be used as part of a higher-level module. When defining one, users should ensure that Terraform is aware of the `juju_model` dependency of the charm module. There are two options to do so when creating a high-level module:
+
+### Define a `juju_model` resource
+Define a `juju_model` resource and pass to the `model_name` input a reference to the `juju_model` resource's name. For example:
+
+```
+resource "juju_model" "testing" {
+  name = kubeflow
+}
+
+module "training-operator" {
+  source = "<path-to-this-directory>"
+  model_name = juju_model.testing.name
+}
+```
+
+### Define a `data` source
+Define a `data` source and pass to the `model_name` input a reference to the `data.juju_model` resource's name. This will enable Terraform to look for a `juju_model` resource with a name attribute equal to the one provided, and apply only if this is present. Otherwise, it will fail before applying anything.
+```
+data "juju_model" "testing" {
+  name = var.model_name
+}
+
+module "training-operator" {
+  source = "<path-to-this-directory>"
+  model_name = data.juju_model.testing.name
+}
+```
--- a/terraform/README.md
+++ b/terraform/README.md
@@ -19,7 +19,7 @@ The module offers the following configurable inputs:
 | `channel`| string | Channel that the charm is deployed from | False |
 | `config`| map(string) | Map of the charm configuration options | False |
 | `model_name`| string | Name of the model that the charm is deployed on | True |
-| `resources`| map(number) | Map of charm resources revisions | False |
+| `resources`| map(string) | Map of the charm resources | False |
 | `revision`| number | Revision number of the charm name | False |
 
 ### Outputs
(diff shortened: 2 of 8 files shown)


.github/workflows/integrate.yaml
```diff
@@ -35,6 +35,14 @@ jobs:
           path: training-operator-cov_html
         if: failure()
 
+  terraform-checks:
+    name: Terraform
+    uses: canonical/charmed-kubeflow-workflows/.github/workflows/terraform-checks.yaml@main
+    with:
+      charm-path: .
+      model: kubeflow
```
DnPlas, line 43: The training operator does not need a model, so you can leave the default one.

DnPlas (commented):
Thanks @orfeas-k, left a comment. Similar to your PR for notebook operators, I think it would be beneficial if we set the default channel to `latest/edge` in the re-usable wf. Let me know your decision and if you change this PR to reflect that change, as that was the only thing I noticed.

terraform/README.md
```diff
@@ -0,0 +1,63 @@
+# Terraform module for training-operator
+
+This is a Terraform module facilitating the deployment of the training-operator charm, using the [Terraform juju provider](https://github.com/juju/terraform-provider-juju/). For more information, refer to the provider [documentation](https://registry.terraform.io/providers/juju/juju/latest/docs). 
+
+## Compatibility
+This terraform module is compatible with charms of version >= 1.8 due to changes in the charm's relations.
```
DnPlas, line 6: Maybe we don't need to provide the whole context. As a reader, if I don't know what relations were in 1.7, I'd probably won't care about this. Perhaps rephrase it to:

```
This terraform module is compatible with the training-operator charm 1.8/stable.
```

I'm hesitant of mentioning any other version as we don't know what comes in the future. It's also going to be an interesting challenge as we generate different versions of a module. Perhaps what we could do is assume that the tf module that lives on the GH branch is compatible only with the charm in the same branch.

.github/workflows/integrate.yaml
```diff
@@ -35,6 +35,14 @@ jobs:
...
+      channel: latest/edge
```
DnPlas, line 44: After reviewing the README, I wonder if we should deploy latest/edge when we say this version of the module is compatible with 1.8. My main concern with that is that we are testing a version but documenting another one, which may be inconsistent. Right now there are no differences between 1.8/edge, 1.8/stable and latest/edge, but if I introduce a new relation tomorrow, then the module won't be compatible with 1.8 anymore, and we'll only be testing latest/edge. wdyt about this scenario?

DnPlas (commented):
Thanks @orfeas-k, few comments that may apply to other PRs as well.

terraform/README.md
orfeas-k: Keeping the module-charm compatibility in the same branch is an interesting approach. It introduces an initial versioning system too and makes more clear the compatibility pattern. What I 'm worried about though is that:
1. this will introduce the chore of adding modules on every release, since the modules will leave on `track` branches. The scenario of not having new relations is a very likely one for many of our charms and keeping the modules in `main` will mean that we may not have to touch the modules (at least for versioning reasons) for quite a while. 
2. This is a bit scratching the surface of defining the policy of when a module needs to be updated, which we haven't fully defined yet and I 'm afraid that next pulse we may change this if we have a task on this.
3. We have already merged quite a few PRs in `main` :sweat_smile: 

Regarding 1, I think that probably in the future, there will be an automated way of creating those modules since  this looks like something easily automated and maybe maintaining different versions could be done by our CI. But until this is possible, I 'm a bit scared of introducing chores.

.github/workflows/integrate.yaml
orfeas-k: The rationale here is that `latest/edge` is >= 1.8 and since we 're saying that this is compatible with >=1.8, then testing is done with a documented version. I preferred to go with `latest/edge` due to merging this in `main`, and it felt weird to test with a track version.

>but if I introduce a new relation tomorrow, then the module won't be compatible with 1.8 anymore, and we'll only be testing latest/edge. wdyt about this scenario?

If we introduced a new relation tomorrow (I guess in `main`), then that would create the need for a new module. Now I 'm not sure how we would handle this exactly tbh. Would we created a new 1.8-compatible in `track/1.8` and update the one in `main` to be latest/edge-compatible? Would we keep only 1.8-compatible in `main`? It does feel a bit weird and is actually tied to the dicsussion comment above.

orfeas-k:
@DnPlas https://github.com/canonical/charmed-kubeflow-workflows/pull/67

terraform/README.md
NohaIhab: > The scenario of not having new relations is a very likely one for many of our charms

I think this is really uncertain, it's always a possibility that we need a new relation / deprecate an existing relation. I wouldn't base our versioning system on this assumption.

> We have already merged quite a few PRs in main 😅

If we agree to change this, we can have a small task to update this in the merged PRs.

> there will be an automated way of creating those modules

I fully agree that we can move to automate the versioning chore in the future, but for now seems that it is something we need to maintain.

Another suggestion is to have a less strict wording to explain the actual situation, like
```
This terraform module is compatible with the training-operator charm 1.8 relations. The terraform module can be compatible with >1.8 versions given there are no changes in the relations from 1.8 version.
```
I'm not a big fan of that^ but proposing something that we can build on.

I think mentioning that the relations / the `outputs.tf` file is the one that affects versioning is an important note.

mvlassis: > Perhaps what we could do is assume that the tf module that lives on the GH branch is compatible only with the charm in the same branch.

I think this is the most sensible option regarding version support, although it will introduce duplicate code, it follows the same strategy we already use for the actual charms. Hopefully, an automated approach in the future could help speed up the whole process.

> Maybe we don't need to provide the whole context. As a reader, if I don't know what relations were in 1.7, I'd probably won't care about this. 

Regarding this, I think with @DnPlas  in that including potential future versions with `>=` in the compatibility section doesn't make much sense, we should only mention the exact version we have tested on. Including older versions of charms (before 1.8) in the compatibility isn't sensible either, since we don't officially support and never test on them. As a general template, I think this is the most precise:
```
This Terraform module is compatible with the training-operator charm 1.8/stable.
```

As supplemental information, we *could* add a small `Note` below that sentence that is directed at people who know about charms and how they work, mentioning that:
```
Note: The Terraform module may be compatible with other versions of the charm given there are no changes in the relations from version `1.8`.
```

(Also note that the word "Terraform" should always be capitalised per their documentation.

.github/workflows/integrate.yaml
NohaIhab: since this is the `main` branch I agree with @orfeas-k to go with `latest/edge`
If we introduce a new relation in `main`, we must keep consistency with the terraform module, meaning that any new relation should take effect in `outputs.tf`. This is something we have to keep in mind from now on.

DnPlas: Based in our discussion from yesterday, we should remove this and the name of the model and use the defaults from the workflow.

terraform/README.md
DnPlas: As discussed in an online sync (see notes [here](https://docs.google.com/document/d/1VYSCHlfvbGeOKgICOIxZkKTFTdLvll_6BTL8xyGyfGA/edit)), we'll go with:

1. Handling module compatibility in branches, that is, the module in Github branch `track/1.8` of the training operator is compatible with that charm (1.8/stable, 1.8/edge), but the module in branch `main` is compatible with that charm (latest/edge).
2. Removing the note from the README.md as it will be obviated that there is a 1:1 compatibility mapping between the module and the charm in the same Github branch

@orfeas-k, may I ask you to do the appropriate changes?

DnPlas (changes requested):
@orfeas-k just a couple of changes, other than that lgtm!

mvlassis (approved):
Nice work!

PR #214, opened 2025-02-04
chore: upgrade training-operator 1.8.0 -> 1.9.0

DnPlas:
This commit upgrades the template files that are used to deploy the training-operator. It also adds an example JAX job for it to be tested.


mvlassis (dismissed):
I went through the process of `diff`ing the manifests and I get the same results. 

Note that there is an issue with the testing of the newly created jaxjob

DnPlas:
> I went through the process of `diff`ing the manifests and I get the same results.
> 
> Note that there is an issue with the testing of the newly created jaxjob

I believe Jaxjobs are missing in the aggregation clusterroles for kubeflow, I have created an issue https://github.com/kubeflow/training-operator/issues/2416. We can block merging this PR until we get a reply from the WG.

DnPlas:
@mvlassis ready for another review. The jaxjob check should pass now. Please note that the CI will still fail because of chisme's version. Please review and approve https://github.com/canonical/training-operator/pull/215 first.

mvlassis (dismissed):
Nice!

mvlassis (approved):

PR #271, opened 2025-10-29
fix: various CI and charm improvements

dariofaccin:
This PR introduces relevant changes to improve reliability of CI and charm behaviour.

### CI improvements
This PR adds a step before integration tests to preload heavy training images, to avoid juju timeouts in long-running tests (in particular for `deepspeed` runtime).
The PR also fixes an issue in the _Publish_ action: since it is a reusable workflow, it cannot be used as a step in a job. For this reason, it is moved back to standalone job and it leverages the output from a previous job. The output is required when pushing on `main-v2`, otherwise the Charmhub channel won't be selected (as documented [here](https://github.com/canonical/charming-actions/tree/2.6.2/channel#branch-selection)).

### Charm improvements
In some situations, upon charm removal CRDs are removed in an order for which some resources are left; this happens most frequently when `ClusterTrainingRuntime` objects and definitions are removed before any `TrainJob` object referencing them: see the [finalizer](https://github.com/kubeflow/trainer/pull/2608) PR on upstream. For this reason, I split the CRD in two files: one for training runtimes (and `JobSet` related stuff), one for `TrainJob`; they have independent handler and on `remove` hook, first delete `TrainJob` objects and CRDs, and then the rest.
Also, at times deletion of `ClusterTrainingRuntimes` takes longer than removal of deployment, causing them to remain hanging.
This PR adds two sanity checks to ensure `TrainJob` and `ClusterTrainingRuntimes` resources are properly deleted before moving on next deletion.

### DIFF
--- a/.github/workflows/integrate.yaml
+++ b/.github/workflows/integrate.yaml
@@ -81,14 +81,23 @@ jobs:
         run: |
           # this step is required to prefetch heavy images to avoid false-positive failures in
           # integration tests due to timeouts unrelated to the charm itself
-          IMAGES=(
-              "ghcr.io/kubeflow/trainer/deepspeed-runtime:latest"
+          RUNTIME_TEMPLATE_FILES=(
+            "src/training_runtimes/deepspeed_distributed.yaml.j2"
           )
-          for image in "${IMAGES[@]}"; do
-              echo "Pulling $image"
-              # explaination of this command can be found in Canonical K8s image management doc
-              # https://documentation.ubuntu.com/canonical-kubernetes/latest/snap/howto/image-management/
-              sudo /snap/k8s/current/bin/ctr --namespace k8s.io image pull "$image" > /dev/null 2>&1
+          for TPL in "${RUNTIME_TEMPLATE_FILES[@]}"; do
+            IMAGE=$(yq '.spec.template.spec.replicatedJobs.[0].template.spec.template.spec.containers[0].image' "${TPL}")
+            # set internal field separator to ":" and tokenize the image string
+            IFS=':' read -ra SPLIT_IMAGE <<< "${IMAGE}"
+            HASH="latest"
+            # handle case in which image already has an hash
+            if [ ${#SPLIT_IMAGE[@]} = 2 ]; then
+              HASH=${SPLIT_IMAGE[-1]}
+            fi
+            FINAL_IMAGE="${SPLIT_IMAGE[0]}:${HASH}"
+            echo "Pulling ${FINAL_IMAGE}"
+            # explanation of this command can be found in Canonical K8s image management doc
+            # https://documentation.ubuntu.com/canonical-kubernetes/latest/snap/howto/image-management/
+            sudo /snap/k8s/current/bin/ctr --namespace k8s.io image pull "${FINAL_IMAGE}" > /dev/null 2>&1
           done
         if: ${{ matrix.tox-environments == 'integration' }}
 


.github/workflows/integrate.yaml
```diff
@@ -76,6 +76,20 @@ jobs:
...
+        id: cache-images
+        shell: bash
+        run: |
+          IMAGES=(
+              "ghcr.io/kubeflow/trainer/deepspeed-runtime:latest"
+          )
+
+          for image in "${IMAGES[@]}"; do
+              echo "Pulling $image"
+              sudo /snap/k8s/current/bin/ctr --address /run/containerd/containerd.sock --namespace k8s.io image pull "$image" > /dev/null 2>&1
```
MattiaSarti: suggestion (non-blocking): shall we add some comment to help understand the nuances of this complex command, for developers in the future to understand it at first glance? Perhaps linking something from [Canonical K8s' docs](https://documentation.ubuntu.com/canonical-kubernetes/latest/snap/howto/image-management/)?

tests/unit/test_charm.py
```diff
@@ -138,13 +153,39 @@ def test_blocked_on_apierror_on_crd_resource_handler(
...
+    def test_blocked_on_apierror_on_trainjob_resource_handler(
+        self,
+        _: MagicMock,
+        __: MagicMock,
+        trainjob_resource_handler: MagicMock,
+        ___: MagicMock,
+        harness: Harness,
+    ):
+        # Ensure the unit is in BlockedStatus
+        # on exception when creating auth resources
```
MattiaSarti: nitpick: the comment should not refer to "auth" resources

src/charm.py
```diff
@@ -254,10 +286,21 @@ def _on_upgrade(self, _):
...
+            delete_many(
+                self.trainjob_resource_handler.lightkube_client, trainjob_resources_manifests
+            )
+            self.ensure_crd_is_deleted(
+                self.trainjob_resource_handler.lightkube_client, "trainjobs.trainer.kubeflow.org"
+            )
             delete_many(self.crd_resource_handler.lightkube_client, crd_resources_manifests)
+            self.ensure_crd_is_deleted(
+                self.crd_resource_handler.lightkube_client,
+                "clustertrainingruntimes.trainer.kubeflow.org",
```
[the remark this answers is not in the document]
dariofaccin: There are two CRDs: `ClusterTrainingRuntimes` for cluster-scoped runtimes, and `TrainingRuntimes` for namespace-scoped runtimes. We don't (and upstream doesn't also) have any of the latter.

.github/workflows/integrate.yaml
dariofaccin: Added a couple of comments on why this is required and reference to documentation for the `ctr` command.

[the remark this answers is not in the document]
dariofaccin: Removed, as it works anyway.

src/charm.py
```diff
@@ -254,10 +286,21 @@ def _on_upgrade(self, _):
...
         self.unit.status = MaintenanceStatus("Removing K8S resources")
+        trainjob_resources_manifests = self.trainjob_resource_handler.render_manifests()
         k8s_resources_manifests = self.k8s_resource_handler.render_manifests()
         crd_resources_manifests = self.crd_resource_handler.render_manifests()
         try:
+            delete_many(
+                self.trainjob_resource_handler.lightkube_client, trainjob_resources_manifests
+            )
+            self.ensure_crd_is_deleted(
+                self.trainjob_resource_handler.lightkube_client, "trainjobs.trainer.kubeflow.org"
```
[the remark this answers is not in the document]
dariofaccin: I split the CRDs in three files: one for `TrainJob` related defs, one for `TrainingRuntimes` related defs and one for `JobSet` related defs. This allows better granularity in sanity checks and also to read the names from the "templates".

tests/unit/test_charm.py
dariofaccin: Fixed this one and other typos.

src/charm.py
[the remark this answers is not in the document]
MattiaSarti: Fair enough, I see the other side of the coin.

.github/workflows/integrate.yaml
```diff
@@ -76,6 +76,22 @@ jobs:
           sudo snap install concierge --classic
           sudo concierge prepare --trace
 
+      - name: Cache images locally
+        shell: bash
+        run: |
+          # this step is required to prefetch heavy images to avoid false-positive failures in
+          # integration tests due to timeouts unrelated to the charm itself
+          IMAGES=(
+              "ghcr.io/kubeflow/trainer/deepspeed-runtime:latest"
```
misohu: question: where is this image comming from ? why is it latest ? how we gonna maintain this line in the future ?

dariofaccin: The image is coming from the [upstream manifest](https://github.com/kubeflow/trainer/blob/486978d2527d9d064e0fda3d62d04a4c45b95d31/manifests/base/runtimes/deepspeed_distributed.yaml#L34). I had to explicitly add the `:latest` suffix otherwise `ctr` fails, but in K8s it always pulls the latest when specified like in the manifest.
@MattiaSarti this could be actually done, but in that case I would iterate over all images for the runtimes; this is not possible, at least using these runners, as the step fails with `no space left on device`.

misohu: Synced with @dariofaccin we agreed to load the image from the manifest file with a script.

PR #275, opened 2025-11-17
feat: add Leader Worker Set and upgrade Trainer/JobSet manifests

dariofaccin:
This PR updates manifests for Trainer, JobSet and adds manifests for the new Leader Worker Set to the charm.


src/templates/lws-service_manifests.yaml.j2
```diff
@@ -0,0 +1,37 @@
...
+    app.kubernetes.io/instance: webhook-service
+    app.kubernetes.io/name: {{ app_name }}-service
+    app.kubernetes.io/part-of: lws
+  name: {{ app_name }}-lws-webhook-service
+  namespace: {{ namespace }}
+spec:
+  ports:
+  - port: 443
+    protocol: TCP
+    targetPort: 9443
```
mvlassis: suggestion(non-blocking): Although I don't see much utility in this, let's use the context values of `{{ webhook_port }}` and `{{ webhook_target_port }}` for consistency, since we're doing the same thing in `jobset-service.yaml`
```suggestion
  - port: {{ webhook_port }}
    protocol: TCP
    targetPort: {{ webhook_target_port }}
```

mvlassis (approved):
Nice work!

PR #282, opened 2026-01-12
feat: run workload and charm as unprivileged user

### LINKED ISSUE #276: Add unprivileged configuration and tests for the Training operator charm
### Context

As a necessary security measure, Charmed Kubeflow is supposed to run all charms and their respective workloads as non-root users in order to prevent threats that come with root privileges.

Assuming the [exploration](https://github.com/canonical/bundle-kubeflow/issues/1354) is completed, charm needs to be configured to run rootless and proper tests need to be implemented to ensure correct privileges.

Ref: https://github.com/canonical/bundle-kubeflow/issues/1368

### What needs to get done

1. Modify charm to run rootless
2. Add proper tests to ensure correct privileges

### Definition of Done

1. Charm is run rootless
2. Workload is run rootless
3. CI is passing

dariofaccin:
This PR introduces rootless platform-component workload and charms.
The charm container is run with user `juju`, with static ID `170`.
The PR includes integration tests to assert that the security context of each container in the pod is correct.

### Custom testing approach
This charm creates a new deployment, usually named `training-operator`, which creates a new pod with one container. This pod is not managed by juju, so it not not possible to fetch its state using the standard approach used for other charms.
For this reason, this charm includes a custom testing setup.
The function `build_pod_container_map` builds a nested map of all charm pods (in this case, it includes only the `charm` container) and deployment pods, taking the container configuration (name, security context) from a user-defined template. This map is then iterated in the test.
The test does not stop on the first failed assertion, but checks all pods and containers and then asserts that none of them is non-compliant. The assertion message includes all pods failing the assessment criteria.

This PR closes #276.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -2249,22 +2249,22 @@ files = [
 
 [[package]]
 name = "protobuf"
-version = "6.33.3"
+version = "6.33.4"
 description = ""
 optional = false
 python-versions = ">=3.9"
 groups = ["charm", "integration"]
 files = [
-    {file = "protobuf-6.33.3-cp310-abi3-win32.whl", hash = "sha256:b4046f9f2ede57ad5b1d9917baafcbcad42f8151a73c755a1e2ec9557b0a764f"},
-    {file = "protobuf-6.33.3-cp310-abi3-win_amd64.whl", hash = "sha256:1fd18f030ae9df97712fbbb0849b6e54c63e3edd9b88d8c3bb4771f84d8db7a4"},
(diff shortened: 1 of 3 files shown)


MattiaSarti (changes requested):
suggestion (blocking): testing the Deployment as described is completely fine, but shouldn't we also explicitly set [its securityContext](https://github.com/canonical/training-operator/blob/main/src/templates/deployment.yaml.j2#L59), to make sure it is not possible to have it somehow overridden if something changes in the future?

PR #293, opened 2026-02-11
feat: integrate kubeflow-trainer with ambient

### LINKED ISSUE #283: Integrate Training Operator with Ambient mesh
### Context

As part of integrating our solution with the ambient mesh charms, we'll need to integrate the Training Operator charm with ambient mesh.

Due to this charm owning a webhook, we have to create an allow-all policy for it to be able to talk to the K8s Controller.

### What needs to get done

1. Integrate the charm with `istio-beacon-k8`
2. Extend the charm to create the allow-all `AuthorizationPolicy`
    * This should be done only if related with `istio-beacon-k8s`
3. Add tests for ambient integration

### Definition of Done

Training Operator can be integrated with ambient mesh

NohaIhab:
Closes #283


pyproject.toml
```diff
@@ -54,6 +54,8 @@ lightkube-models = "^1.31.1.8"
 ops = "^2.17.1"
 pyyaml = "^6.0.2"
 tenacity = "^9.0.0"
+lightkube-extensions = "^0.3.0"
+charmed-service-mesh-helpers = "^0.5.0"
```
MattiaSarti, line 58: note: these changes should not be necessary if you run `tox -e update-requirements` because the `poetry.lock` file would be updated with the latest compatible versions and they would be included because the latest Chisme requires them (as proof, they are indeed already part of [the lock file on `main-v2`](https://github.com/canonical/training-operator/blob/main-v2/poetry.lock) even without having ever explicitly required them before), but I am completely fine with keeping them to cover the need in case we somehow decide to get rid of Chisme in the future

tox.ini
MattiaSarti, line 121: suggestion (blocking): shouldn't we also have an ambient-mode version of these integration tests with Profiles where both TrainerV2 and Profiles are run in ambient mode?

src/charm.py
```diff
@@ -226,6 +257,78 @@ def trainjob_resource_handler(self):
...
+    def _reconcile_policy_resource_manager(self):
+        """Reconcile authorization policies via PolicyResourceManager."""
+        if self._mesh._relation:
+            self._policy_resource_manager.reconcile(
+                policies=[], mesh_type=self._mesh.mesh_type, raw_policies=self._allow_all_policies
+            )
+
+    def _remove_authorization_policies(self, _):
+        """Remove authorization policies via PolicyResourceManager."""
+        self._policy_resource_manager.delete()
```
MattiaSarti: suggestion (blocking): I believe here we need a leadership check or conditional, because the library does not seem to handle policy updates only with leader units

src/charm.py
```diff
@@ -226,6 +257,78 @@ def trainjob_resource_handler(self):
...
+                spec=AuthorizationPolicySpec(
+                    selector=WorkloadSelector(
+                        matchLabels={
+                            "app.kubernetes.io/name": f"{app_name}-{component}",
+                        },
+                    ),
+                    action=Action.allow,
+                    rules=[Rule()],
+                ).model_dump(by_alias=True, exclude_unset=True, exclude_none=True),
+            )
```
MattiaSarti, line 317: praise: very maintainable generalization on our labels 🎩

tests/unit/test_charm.py
```diff
@@ -286,3 +299,132 @@ def test_on_remove_failure(
...
+        self,
+        service_mesh_consumer_cls: MagicMock,
+        policy_resource_manager_cls: MagicMock,
+        policy_client: MagicMock,
+        delete_many: MagicMock,
+        _: MagicMock,
+        __: MagicMock,
+        ___: MagicMock,
+        ____: MagicMock,
+        _____: MagicMock,
```
MattiaSarti: nitpick: all but `policy_resource_manager_cls` can be `__...`

MattiaSarti (changes requested):
Excellent work!
I have a couple of blocking concerns, though:

pyproject.toml
NohaIhab: keeping them because we are using the lib directly, so as you said it'd be an issue if we were to remove them from chisme

tox.ini
NohaIhab: I did not replicate this one because the tests there are just:
1. Deploying training and profiles
2. Checking on the RBAC permissions for the kubeflow profile on the training resources

The only reason istio pilot is deployed there is so we can get the `AuthorizationPolicy` CRD which is surely also available from istio k8s. I don't see value in adding the charms to the mesh for this test case. WDYT?

src/charm.py
NohaIhab: https://github.com/canonical/training-operator/pull/293/commits/6f053b22530a06ebbe272f0f968c939d187b292b

tests/unit/test_charm.py
NohaIhab: https://github.com/canonical/training-operator/pull/293/commits/b456dc7f1c39f2932242cf2d129c0e99da3c899d

tox.ini
MattiaSarti: I see your point, so we can postpone the update of those tests. Hopefully, then, when migrating completely to ambient mode, it will be straightforward to update those tests.

MattiaSarti (approved):
Excellent!

kimwnasptd (approved):

PR #295, opened 2026-02-18
fix: do not add selector labels to allow seamless upgrade

### LINKED ISSUE #294: Cannot upgrade `training-operator` from `1.9/stable` to `1.9/edge`
### Bug Description

Upgrade from `1.9/stable` to `1.9/edge` fails because we updated the selector labels in the deployment manifest.

### To Reproduce

1. `juju deploy training-operator --channel=1.10/stable`
2. `juju refresh training-operator --channel=1.10/edge`

### Environment

N/A

### Relevant Log Output

```shell
lightkube.core.exceptions.ApiError: Deployment.apps "training-operator" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"app.kubernetes.io/name":"training-operator-manager", "control-plane":"kubeflow-training-operator"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable
```

### Additional Context

_No response_

dariofaccin:
This PR removes the newly added label `app.kubernetes.io/name` from the `Deployment`.
This change is required since labels are immutable and therefore the charm upgrade fails.
An additional helper function, named `get_deployment_pod_names`, is required to retrieve the pod matching the `control-plane` label.
Label selector required for Service Mesh is updated to use the `control-plane` label, which only matches the training manager (not the charm agent pod).
Using such label, we are able to identify only the workload pod:
```
kubectl get pods -A -l "control-plane=kubeflow-training-operator"
NAMESPACE   NAME                                 READY   STATUS    RESTARTS   AGE
kubeflow    training-operator-6f8d557494-rbsjq   1/1     Running   0          151m
```
The charm pod is instead matched with the `app.kubernetes.io/name` label:
```
kubectl get pods -A -l "app.kubernetes.io/name=training-operator" 
NAMESPACE   NAME                  READY   STATUS    RESTARTS   AGE
kubeflow    training-operator-0   1/1     Running   0          154m
```

This PR closes #294.

### DIFF
--- a/.github/workflows/integrate.yaml
+++ b/.github/workflows/integrate.yaml
@@ -64,7 +64,7 @@ jobs:
           - integration-with-profiles
     steps:
       - name: Maximise GH runner space
-        uses: jlumbroso/free-disk-space@v1.3.1
+        uses: jlumbroso/free-disk-space@54081f138730dfa15788a46383842cd2f914a1be
 
       - name: Check out code
         uses: actions/checkout@v4
--- a/.github/workflows/integrate.yaml
+++ b/.github/workflows/integrate.yaml
@@ -21,6 +21,7 @@ jobs:
     steps:
     - name: Check out code
       uses: actions/checkout@v4
+
     - name: Install dependencies
       run: pipx install tox
 
@@ -85,8 +86,18 @@ jobs:
           kubectl -n kube-system patch configmap cilium-config --type merge --patch '{"data":{"bpf-lb-sock-hostns-only":"true"}}'
           kubectl -n kube-system rollout restart daemonset cilium
   
+      - name: Fetch charm
+        uses: actions/download-artifact@v5
+        with:
+          name: built-charm
+          path: built/
+
+      - name: Get charm path
+        id: charm-path
+        run: echo "charm_path=$(find built/ -name '*.charm' -type f -print)" >> $GITHUB_OUTPUT
+
       - name: Run integration tests
-        run: tox -e ${{ matrix.tox-environment }} -- --model testing
+        run: tox -e ${{ matrix.tox-environment }}  -- --model testing --charm-path="${{ steps.charm-path.outputs.charm_path }}"
 
       - name: Capture k8s resources on failure
         run: |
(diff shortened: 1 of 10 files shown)


tests/integration/test_charm.py
```diff
@@ -108,7 +109,7 @@ async def ensure_training_operator_is_running(ops_test: OpsTest) -> None:
         "status.phase!=Running",
         check=True,
     )
-    assert "training-operator" not in out
+    assert APP_NAME not in out
```
MattiaSarti, line 115: praise: thanks for also making this tidier along the way

src/charm.py
```diff
@@ -196,7 +196,7 @@ def generate_allow_all_authorization_policy(
             spec=AuthorizationPolicySpec(
                 selector=WorkloadSelector(
                     # Use the unique label from src/templates/deployment.yaml.j2
-                    matchLabels={"app.kubernetes.io/name": f"{app_name}-manager"},
+                    matchLabels={"control-plane": f"{namespace}-{app_name}"},
```
MattiaSarti, line 199: note: as noted down with @dariofaccin during development, this uniquely identifies the same workload and the allow-all policy still matches it and only it

MattiaSarti (approved):

src/charm.py
dariofaccin: I updated the PR description to make it more clear with an example taken from a live deployment.

.github/workflows/integrate.yaml
```diff
@@ -85,8 +86,18 @@ jobs:
...
+          name: built-charm
+          path: built/
+
+      - name: Get charm path
+        id: charm-path
+        run: echo "charm_path=$(find built/ -name '*.charm' -type f -print)" >> $GITHUB_OUTPUT
+
       - name: Run integration tests
-        run: tox -e ${{ matrix.tox-environment }} -- --model testing
+        run: tox -e ${{ matrix.tox-environment }}  -- --model testing --charm-path="${{ steps.charm-path.outputs.charm_path }}"
```
NohaIhab: praise: thank you for the CI improvement

.github/workflows/integrate.yaml
NohaIhab, line 1: nit: I think `Get operator logs` step here needs to be updated

NohaIhab (approved):
Thanks @dariofaccin

.github/workflows/integrate.yaml
dariofaccin: Updated the steps with correct labels for the operator and also added steps for the workload pod.

NohaIhab (approved):

PR #305, opened 2026-05-22
chore: Update CI to use charmcraftcache

mvlassis:
## Summary
This PR updates the CI to utilize charmcraftcache, by using actions from [data-platform-workflows](https://github.com/canonical/data-platform-workflows). The ci.yaml is very similar to the structure we have in our multi-repo charms, for example see https://github.com/canonical/knative-operators/blob/main/.github/workflows/ci.yaml.

Base issue: https://github.com/canonical/bundle-kubeflow/issues/1429


dariofaccin (approved):

.github/workflows/ci.yaml
```diff
@@ -78,11 +168,12 @@ jobs:
           # See https://canonical-service-mesh-documentation.readthedocs-hosted.com/en/latest/how-to/use-charmed-istio-with-canonical-kubernetes/
           kubectl -n kube-system patch configmap cilium-config --type merge --patch '{"data":{"bpf-lb-sock-hostns-only":"true"}}'
           kubectl -n kube-system rollout restart daemonset cilium
```
MattiaSarti: suggestion (non-blocking): the link is not found and the commands may also be missing an additional instruction that it could be beneficial to include, so I would recommend using a different version of this section that can be found [here](https://github.com/canonical/istio-operators/blob/main/.github/workflows/ci.yaml#L142-L145)

MattiaSarti (approved):

.github/workflows/ci.yaml
mvlassis: Should be updated in https://github.com/canonical/training-operator/pull/305/commits/edbbfc9bb61aa90df1069c81a8c74a726763e4a0

dariofaccin (changes requested):
issue(blocking): we need to update also `on_pull_request` workflow not to reference old ones.

mvlassis:
I removed the unused jobs in `on_pull_request.yaml` in https://github.com/canonical/training-operator/pull/305/commits/f443a720f76a6fcc8d3dadbe21a802734afddaf8

dariofaccin (approved):

